PR #1, opened 2019-09-16
Build alloc list

jacob-hughes:
Store the size and pointer of each allocation

In order for a conservative collector to trace the object graph it must know the following two things about each potential root on the stack:
    1 The size of an allocated pointer's block
    2 How to get to the base pointer from a pointer inside an allocation
      block

This PR introduces a separate array which records this metadata on each allocation. It uses the allocator in libc and provides an additional API for a collector to query information about pointers at runtime.

In this PR, the allocation list is set arbitrarily at 2KB, and will abort when full. A PR to fix this so that the list grows dynamically will be forthcoming. Each metadata entry in the allocation list is 2
machine words and includes the allocated pointer, and the block size.

Insertion into the allocation list is done with an O(1) pointer bump until it is fully initialised, at which point it becomes O(n) looking for freed blocks as the list becomes fragmented. Removal of each entry
is O(n).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "gcmalloc"
+version = "0.1.0"
+authors = ["Jacob Hughes <jh@jakehughes.uk>"]
+edition = "2018"
+
+[dependencies]
+libc = "*"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,358 @@
+// Copyright (c) 2019 King's College London
+// Created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any
+// person obtaining a copy of this software, associated documentation and/or
+// data (collectively the "Software"), free of charge and under any and all
+// copyright rights in the Software, and any and all patent rights owned or
+// freely licensable by each licensor hereunder covering either (i) the
+// unmodified Software as contributed to or provided by such licensor, or (ii)
+// the Larger Works (as defined below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software
+// is contributed by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create
+// derivative works of, display, perform, and distribute the Software and make,
+// use, sell, offer for sale, import, export, have made, and have sold the
+// Software and the Larger Work(s), and to sublicense the foregoing rights on
+// either these or other terms.
+//
+// This license is subject to the following condition: The above copyright
+// notice and either this complete permission notice or at a minimum a reference
+// to the UPL must be included in all copies or substantial portions of the
+// Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#![crate_name = "gcmalloc"]
+#![crate_type = "rlib"]
+#![feature(core_intrinsics)]
+// Allocators are not allowed to depend on the standard library which in turn
+// requires an allocator in order to avoid circular dependencies. This crate,
+// however, can use all of libcore.
+#![feature(no_std)]
+#![no_std]
+
+// Our system allocator will use the in-tree libc crate for FFI bindings. Note
+// that currently the external (crates.io) libc cannot be used because it links
+// to the standard library (e.g. `#![no_std]` isn't stable yet), so that's why
+// this specifically requires the in-tree version.
+
+extern crate libc;
+
+static SIZE_ALLOC_TABLE: usize = (1024 * 1024) * 2; // 2KB
+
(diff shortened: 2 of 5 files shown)


ltratt:
> it becomes O(n) looking for freed blocks as the list becomes fragmented. Removal of each entry is O(n).

I might be missing something, but that seems strange. If you've got O(n) deletion, I assume it's because you compact as you delete, but that would make adding entries O(1)? Or, conversely, if adding entries is O(n), then deletion doesn't compact and is O(1)?

Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "gcmalloc"
+version = "0.1.0"
+authors = ["Jacob Hughes <jacob.hughes@kcl.ac.uk>"]
```
ltratt: I suggest you use a non-KCL email address that you're more likely to monitor in the long term, but it's up to you.

src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+#![no_std]
+
+// Our system allocator will use the in-tree libc crate for FFI bindings. Note
+// that currently the external (crates.io) libc cannot be used because it links
+// to the standard library (e.g. `#![no_std]` isn't stable yet), so that's why
+// this specifically requires the in-tree version.
+
+extern crate libc;
+
+static SIZE_ALLOC_TABLE: usize = (1024 * 1024 * 1024) * 2; // 2KB
```
ltratt: I assume that the units here are number of entries, not number of kilobytes? If it's the latter, isn't this actually 2MiB?

src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+impl<'a> Iterator for AllocListIterMut<'a> {
+    type Item = &'a mut Block;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // It's UB to call `.add` on a pointer past its allocation bounds, so we
+        // need to check that it's within range before turning it into a pointer
+        // and dereferencing it
+        let ulimit = self.alloc_list.start as usize + SIZE_ALLOC_TABLE;
```
ltratt: Might this check be easier as something like:

```rust
if self.index * size_of::<...>() >= SIZE_ALLOC_TABLE {
    None
} else {
    unsafe { Some(...)}
}
```

This way you remove the (very small chance) that you overflow `ptr`.

src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+                } else {
+                    self.can_bump = false;
+                }
+            }
+        }
+
+        // Slow path, we need to linearly scan for the next free block in the
+        // heap.
+        for block in self.iter_mut() {
+            match block {
```
ltratt: If you have a `_ => ()` case, you probably want `if let` (and it reduces indentation by 1 level, which is nice!).

src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+                }
+                _ => (),
+            }
+        }
+    }
+
+    /// Given an arbitrary pointer (base or inner), finds pointer info of the
+    /// associated base pointer if it exists.
+    fn find_base(&self, ptr: usize) -> Option<PtrInfo> {
+        for block in self.iter() {
```
ltratt: Might it be better to use `self.iter().position(...)` as I think the compiler can generally better optimise that than a`for` loop?

src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+
+static mut PTR_METADATA: Option<AllocList> = None;
+
+fn abort() {
+    unsafe { core::intrinsics::abort() };
+}
+
+/// ============================================================================
+/// Listed below are the five allocation functions currently required by custom
+/// allocators. Their signatures and symbol names are not currently typechecked
```
ltratt: The sentence starting "their signatures" doesn't quite parse.

Cargo.toml
jacob-hughes: Good idea

src/lib.rs
jacob-hughes: That's much nicer! It would need to be `self.index + 1 * ...` as it's 0-indexed.

src/lib.rs
jacob-hughes: Fixed

src/lib.rs
jacob-hughes: Yes, what a stupid mistake.

jacob-hughes:
> If you've got O(n) deletion, I assume it's because you compact as you delete, but that would make adding entries O(1)?

It's much dumber than you think - no compaction on deletion. Insertion is O(n) because once the list has been fully initialised with fast pointer bumping, the list must be searched looking for an available `Block::Free`. It's O(n) deletion because you have to search for the value before you delete it (you cannot assume indexed order). In essence, like malloc, it will fragment over time.

src/lib.rs
jacob-hughes: This comment was copied from the allocator API documentation, I admit it doesn't make a great deal of sense to me either so I've removed it.

jacob-hughes:
I've addressed your comments and ready to squash if you're happy.

src/lib.rs
ltratt: Ah, maybe then just `find` or equivalent. Basically the compiler has a better chance, in general, of optimising those calls then it does a generic `for` loop.

ltratt:
> It's O(n) deletion because you have to search for the value before you delete it (you cannot assume indexed order).

Ah, I see.

src/lib.rs
```diff
@@ -87,15 +87,13 @@ impl<'a> Iterator for AllocListIterMut<'a> {
         // It's UB to call `.add` on a pointer past its allocation bounds, so we
         // need to check that it's within range before turning it into a pointer
         // and dereferencing it
-        let ulimit = self.alloc_list.start as usize + SIZE_ALLOC_TABLE;
-        let ptr = self.alloc_list.start as usize + (self.index * core::mem::size_of::<Block>());
-
-        self.index += 1;
-
-        if ptr >= ulimit {
+        if self.index * core::mem::size_of::<Block>() >= SIZE_ALLOC_TABLE {
```
ltratt: Random (shallow) comment: would `index` be more idiomatic as `idx`?

jacob-hughes: Fixed

src/lib.rs
jacob-hughes: The new approach uses `find_map`.

ltratt:
Works for me. I suggest you add a Travis config and use the softdev rustfmt settings in a new commit. Then we can squash after that's agreed upon.

jacob-hughes:
Added Travis config and Rustfmt config

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #2, opened 2019-09-19
Make allocator thread-safe and switch to Rust's stabilised `GlobalAlloc` API

jacob-hughes:
This PR includes two key changes:

1. Access to the `ALLOC_INFO` list - which contains metadata about each allocation block - is now thread-safe. 

2. Allocation is now done via Rust's [GlobalAlloc](https://doc.rust-lang.org/beta/std/alloc/trait.GlobalAlloc.html) API.

Please see the individual commit messages for more detailed information about each change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,8 +39,55 @@
 #![crate_type = "rlib"]
 #![feature(core_intrinsics)]
 
+use std::alloc::{GlobalAlloc, Layout};
+
+struct GCMalloc;
+
+#[global_allocator]
+static ALLOCATOR: GCMalloc = GCMalloc;
 
 static SIZE_ALLOC_INFO: usize = (1024 * 1024) * 2; // 2KB
+static mut ALLOC_INFO: Option<AllocList> = None;
+
+unsafe impl GlobalAlloc for GCMalloc {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ptr = libc::malloc(layout.size() as libc::size_t) as *mut u8;
+
+        match ALLOC_INFO {
+            Some(ref mut pm) => pm.insert(ptr as usize, layout.size()),
+            None => {
+                let mut al = AllocList::new();
+                al.insert(ptr as usize, layout.size());
+                ALLOC_INFO = Some(al);
+            }
+        };
+        ptr
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        libc::free(ptr as *mut libc::c_void);
+        ALLOC_INFO.as_mut().unwrap().remove(ptr as usize);
+    }
+
+    unsafe fn realloc(&self, ptr: *mut u8, _layout: Layout, new_size: usize) -> *mut u8 {
+        let new_ptr = libc::realloc(ptr as *mut libc::c_void, new_size) as *mut u8;
+
+        if ptr.is_null() {
+            // realloc is equivalent to malloc if called with a null pointer,
+            // thus we must insert the new block info in the ALLOC_INFO list.
+            ALLOC_INFO.as_mut().unwrap().insert(ptr as usize, new_size);
+        } else {
+            // In all other cases, the memory block is resized and a new ptr
+            // is returned. This may differ from the original ptr if the new
+            // block is too large to fit in the same space as the old block
+            ALLOC_INFO
+                .as_mut()
+                .unwrap()
+                .update(ptr as usize, new_ptr as usize, new_size);
+        }
+        new_ptr
+    }
+}
 
 /// A contiguous chunk of memory which records metadata about each pointer
 /// allocated by the allocator. Any pointer can be queried during runtime to
@@ -132,6 +179,9 @@ impl AllocList {
     /// point, insertion is O(n) while it linearly scans the list for the next
     /// free entry.
     fn insert(&mut self, ptr: usize, size: usize) {
+        debug_assert_ne!(ptr, 0);
+        debug_assert_ne!(size, 0);
+
         if self.can_bump {
             unsafe {
                 let next_ptr = self.start.add(self.next_free) as *mut Block;
@@ -189,11 +239,14 @@ impl AllocList {
 
     /// Updates the size associated with a base pointer (perfomed on a realloc).
     /// This must not be called with an inner pointer.
-    fn update(&mut self, ptr: usize, size: usize) {
+    fn update(&mut self, ptr: usize, new_ptr: usize, size: usize) {
         for block in self.iter_mut() {
             if let Block::Entry(base_ptr, _) = *block {
                 if ptr == base_ptr.get() {
-                    *block = Block::Entry(base_ptr, size);
+                    *block = Block::Entry(
+                        unsafe { core::num::NonZeroUsize::new_unchecked(new_ptr) },
+                        size,
+                    );
                     return;
                 }
             }
@@ -224,7 +277,6 @@ struct PtrInfo {
     size: usize,
 }
 
-static mut PTR_METADATA: Option<AllocList> = None;
 
 fn abort() {
     unsafe { core::intrinsics::abort() };
@@ -273,7 +325,7 @@ mod tests {
 
         let num_ptrs = SIZE_ALLOC_INFO / core::mem::size_of::<Block>();
         for i in 0..num_ptrs {
-            al.insert(i, 1);
+            al.insert(i + 1, 1);
         }
 
         // // Free the pointer in the middle of the list
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@
 #![feature(core_intrinsics)]
 
 
-static SIZE_ALLOC_TABLE: usize = (1024 * 1024) * 2; // 2KB
+static SIZE_ALLOC_INFO: usize = (1024 * 1024) * 2; // 2KB
 
 /// A contiguous chunk of memory which records metadata about each pointer
 /// allocated by the allocator. Any pointer can be queried during runtime to
@@ -75,7 +75,7 @@ impl<'a> Iterator for AllocListIterMut<'a> {
         // It's UB to call `.add` on a pointer past its allocation bounds, so we
         // need to check that it's within range before turning it into a pointer
         // and dereferencing it
-        if self.idx * core::mem::size_of::<Block>() >= SIZE_ALLOC_TABLE {
+        if self.idx * core::mem::size_of::<Block>() >= SIZE_ALLOC_INFO {
             return None;
         }
 
@@ -92,7 +92,7 @@ impl<'a> Iterator for AllocListIter<'a> {
     fn next(&mut self) -> Option<&'a Block> {
         // It's UB to call `.add` on a pointer past its allocation bounds, so we
         // need to check that it's within range before turning it into a pointer
-        if self.idx * core::mem::size_of::<Block>() >= SIZE_ALLOC_TABLE {
+        if self.idx * core::mem::size_of::<Block>() >= SIZE_ALLOC_INFO {
             return None;
         }
 
@@ -106,7 +106,7 @@ impl<'a> Iterator for AllocListIter<'a> {
 
 impl AllocList {
     fn new() -> AllocList {
-        let raw = unsafe { libc::malloc(SIZE_ALLOC_TABLE as libc::size_t) } as *mut Block;
+        let raw = unsafe { libc::malloc(SIZE_ALLOC_INFO as libc::size_t) } as *mut Block;
         AllocList {
             start: raw,
             next_free: 0,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -40,7 +40,7 @@
 #![feature(core_intrinsics)]
 
 
-static SIZE_ALLOC_TABLE: usize = (1024 * 1024) * 2; // 2KB
+static SIZE_ALLOC_INFO: usize = (1024 * 1024) * 2; // 2KB
```
ltratt: I still don't think this is 2KiB!

src/lib.rs
```diff
@@ -39,8 +39,48 @@
...
+            // Spin
+        }
+    }
+
+    fn unlock(&self) {
+            ALLOC_LOCK.store(false, Ordering::Release);
+    }
+}
+
+impl Deref for AllocLock {
```
ltratt: Why would anyone deref `AllocLock`?

src/lib.rs
```diff
@@ -229,11 +284,14 @@ impl AllocList {
...
     /// Updates the size associated with a base pointer (perfomed on a realloc).
     /// This must not be called with an inner pointer.
-    fn update(&mut self, ptr: usize, size: usize) {
+    fn update(&mut self, ptr: usize, new_ptr: usize, size: usize) {
         for block in self.iter_mut() {
             if let Block::Entry(base_ptr, _) = *block {
                 if ptr == base_ptr.get() {
-                    *block = Block::Entry(base_ptr, size);
+                    *block = Block::Entry(
+                        unsafe { core::num::NonZeroUsize::new_unchecked(new_ptr) },
```
ltratt: I feel like a `debug_assert_ne!(new_ptr, 0)` or similar might be a good idea, just to check that nobody ever violates the `NonZeroUsize` property (because if they do, bad things can happen).

ltratt:
I have some minor comments but, overall, this looks good.

One additional comment over the inline comments: https://github.com/softdevteam/gcmalloc/pull/2/commits/babf38096f91a95bc5b871028ad914810bca6af4 feels like it's incomplete. It defines code but doesn't use it?

src/lib.rs
jacob-hughes: This just lets you call `.lock()` `.unlock()` on `AllocLock` and have them update the atomic directly. It's just a bit more ergonomic than top level fn declarations.

jacob-hughes: ... But I've just noticed that I've forgotten to make `lock()` and `unlock() reference `self`. Which is why this might have been non-obvious!

ltratt: I'm not sure I understand. Couldn't `AllocLock.lock` have code along the lines of `self.0.compare_and_swap` and then you can remove this `Deref` implementation?

jacob-hughes: Yes. This was more useful when I implemented an RAII guard (similar to `sys::Sync::MutexGuard`) so that `unlock()` was automatically called as part of `AllocLock.drop()`. I removed it though because it felt overkill. As such, I suppose this can go too.

ltratt: Less code is certainly good :)

src/lib.rs
jacob-hughes: nothing to see here... :eyes:

src/lib.rs
jacob-hughes: I think you're right and there's probably not much harm in doing so. In most cases these calls are made by the compiler, so you'd hope they followed their own rules. If we add assertions, I think these checks should probably happen in the actual `alloc`, `dealloc`, `realloc` calls instead. 

Rust makes other assumptions about allocation which would result in UB which are perfectly legal in system `malloc` / `free` / `realloc`. For example, if `libc::realloc` is called with `size = 0`, it acts as `libc::free`. In rust this is UB. I'm guessing they are more restrictive for portability with other allocator implementations.

jacob-hughes:
> One additional comment over the inline comments: [babf380](https://github.com/softdevteam/gcmalloc/commit/babf38096f91a95bc5b871028ad914810bca6af4) feels like it's incomplete. It defines code but doesn't use it?

It doesn't use it because `lock()` and `unlock()` are called in the new `AllocTrait` implementation which is part of the next commit chronologically. In hindsight, it may have made more sense to re-order this.  5578de4 should have happened first followed by babf380 to make it thread-safe. I can do this as part of squashing if you like.

src/lib.rs
ltratt: In general (and it's hard to make a hard rule for this), I think it's not a bad idea to have a `debug_assert` before an `unsafe` block to check that whatever condition is expected to hold really holds. In some senses, this is often as useful as documentation as it is an actual check. In other words, I don't mind having "duplicate" `debug_assert`'s around, because it can help give us confidence about individual `unsafe` blocks even if we don't understand the full system context.

ltratt:
>  5578de4 should have happened first followed by babf380 to make it thread-safe. I can do this as part of squashing if you like.

OK. Once we've worked out what to do with the `debug_assert`'s (if anything), then this sounds like a reasonable thing to do.

src/lib.rs
vext01: He was just testing you Laurie. Well done, you passed 😆

src/lib.rs
vext01: You could also use the `with` pattern, like [`LocalKey`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.with). This'd make the critical section, in effect, scoped and call sites couldn't forget to release the lock.

Could be considered over-engineering though...

vext01:
Do you foresee other uses of this kind of spinlock in the future?

src/lib.rs
```diff
@@ -39,8 +39,48 @@
...
+
+impl AllocLock {
+    fn lock(&self) {
+        while !ALLOC_LOCK.compare_and_swap(false, true, Ordering::AcqRel) {
+            // Spin
+        }
+    }
+
+    fn unlock(&self) {
+            ALLOC_LOCK.store(false, Ordering::Release);
```
vext01: Is this indent right? Looks over-cooked.

src/lib.rs
```diff
@@ -39,8 +39,48 @@
...
+///
+/// Instead, we use a simple global spinlock built on an atomic bool to guard
+/// access to the ALLOC_INFO list.
+static ALLOC_LOCK: AllocLock = AllocLock(AtomicBool::new(false));
+
+struct AllocLock(AtomicBool);
+
+impl AllocLock {
+    fn lock(&self) {
+        while !ALLOC_LOCK.compare_and_swap(false, true, Ordering::AcqRel) {
```
vext01: I think you may have mentioned this, but shouldn't it be:
```
while !self.0.compare_and_swap(...
```

i.e. `self`.

src/lib.rs
```diff
@@ -82,6 +89,54 @@ impl Deref for AllocLock {
     }
 }
 
+unsafe impl GlobalAlloc for GCMalloc {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
```
vext01: Would the actual allocation, and the recording of the metadata ever need to be atomic?

If a thread walked your allocation records while another thread was on line 95, then the first thread would miss an allocation.

Not sure if that's important or not...

src/lib.rs
```diff
@@ -82,6 +89,54 @@ impl Deref for AllocLock {
...
+        let ptr = libc::malloc(layout.size() as libc::size_t) as *mut u8;
+
+        ALLOC_LOCK.lock();
+        match ALLOC_INFO {
+            Some(ref mut pm) => pm.insert(ptr as usize, layout.size()),
+            None => {
```
vext01: So this is lazily initialised. I guess you couldn't figure out a way to have it created at the start?

src/lib.rs
```diff
@@ -82,6 +89,54 @@ impl Deref for AllocLock {
...
+                al.insert(ptr as usize, layout.size());
+                ALLOC_INFO = Some(al);
+            }
+        };
+        ALLOC_LOCK.unlock();
+        ptr
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        libc::free(ptr as *mut libc::c_void);
```
vext01: Similar question here about atomicity.

src/lib.rs
```diff
@@ -229,11 +284,14 @@ impl AllocList {
 
     /// Updates the size associated with a base pointer (perfomed on a realloc).
     /// This must not be called with an inner pointer.
-    fn update(&mut self, ptr: usize, size: usize) {
+    fn update(&mut self, ptr: usize, new_ptr: usize, size: usize) {
         for block in self.iter_mut() {
```
vext01: Possible optimisation for later: store allocations sorted so that locating the correct pointer can be log time using a chop?

src/lib.rs
vext01: > In most cases these calls are made by the compiler

So an assertion could serve as  a check of our understanding of the compiler ;)

vext01:
That was a fun one to review. Many comments :)

src/lib.rs
jacob-hughes: I considered it, but I actually care about faster insertion than lookup. Storing allocations sorted would have worst case O(n^2). Insertion happens on every single `alloc` call, whereas lookup will happen only during the STW phase of GC. 

Arguably it makes sense to try and reduce the STW pause in the future, but at a cost of slowing down every allocation is, I think, the wrong trade-off.

src/lib.rs
jacob-hughes: Yes, I was testing you

src/lib.rs
jacob-hughes: No, I think this is what travis is complaining about

src/lib.rs
jacob-hughes: Probably at this stage, but maybe worth revisiting in the future

src/lib.rs
jacob-hughes: Unfortunately not without it becoming part of `rustc`. Ideally this initialisation would happen before `main`.

src/lib.rs
jacob-hughes: I don't think so. The call to `malloc` itself is thread-safe, so it will ensure that `ptr` gets a unique value each time it's called. If the thread A is preempted on line 95, thread B may insert its pointer into `ALLOC_INFO` before thread A, even if thread B called `alloc` after. Since `ALLOC_INFO` makes no guarantees about ordering, I think this is okay.

src/lib.rs
jacob-hughes: Again, I don't think `dealloc` needs to happen atomically. If the thread is preempted before acquiring the lock, it will just mean that the ptr info is removed from `ALLOC_INFO` after another allocation has finished.

src/lib.rs
vext01: So your example concerns two threads wanting to insert at the same time. I agree this is safe.

But will there ever be a thread which (e.g.) reads the whole list for some reason and requires a *complete* view of the current heap?

src/lib.rs
vext01: ok

ltratt: `lazy_static` and the like do similar tricks, they just make them a bit less visible.

vext01: I think there was a reason he couldn't use `lazy_static` too. What was it again, Jake?

jacob-hughes: I decided against `lazy_static` as it obscured the locking logic.

ltratt: Sorry, just to be clear, I agree with not using `lazy_static` -- I was just pointing out the (rough) similarity.

src/lib.rs
ltratt: I think we need to define precisely when code can access the list of memory stuff and what it can do with it. For example, we can define that it's safe for code to add and remove stuff to `ALLOC_INFO` at will (provided `ALLOC_LOCK` is held) if the only time we actually make use of the stuff in `ALLOC_INFO` is during a stop-the-world GC (which, by definition, is an implicit lock).

However, in contrast, we might allow threads to read from `ALLOC_INFO` outside of a stop-the-world GC at which point a thread might be interrupted after `free` but before `ALLOC_LOCK.lock()`, leading to a possible read-after-free possibility. If that's the case then we would need to do something like:

```rust
ALLOC_LOCK.lock()
free(...)
ALLOC_INFO....remove(...);
ALLOC_LOCK.unlock()
```

I don't have a firm opinion on what the best thing to do here is, but I'm fairly sure we need to explicitly pick one choice.

vext01: Right, Laurie and I essentially have the same query:  "Should the malloc/free's be inside the critical section?"

jacob-hughes: Right now, I can only ever see the need to read from this table during a STW pause, which, as you've mentioned, implicitly locks `ALLOC_INFO`. I think making this assumption is okay for now providing it's well documented on any read method.

vext01: OK.

jacob-hughes:
Addressed review comments and fixed the failing build.

src/lib.rs
ltratt: Let's put an explicit comment about STW somewhere then, so we're clear in this PR, and clear going forward.

jacob-hughes: Done

ltratt:
If @vext01 is OK with everything then I think you can squash.

vext01:
@vext01 approves. Please squash.

jacob-hughes:
Squish

PR #4, opened 2019-09-25
Separate managed GC allocation from standard system allocation

jacob-hughes:
We intend to use GC alongside normal RAII style memory allocation. This means that during a collection, the collector needs to know which blocks in the heap it is responsible for reclaiming.

This PR includes a separate allocator -- `GC_ALLOCATOR` -- intended to be used instead of the global allocator for GC objects. `GC_ALLOCATOR` ensures that the block is recorded inside the `ALLOC_INFO` list as being GC managed. This info can then be queried during a collection.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -66,10 +66,10 @@ unsafe impl GlobalAlloc for GCMalloc {
 
         ALLOC_LOCK.lock();
         match ALLOC_INFO {
-            Some(ref mut pm) => pm.insert(ptr as usize, layout.size()),
+            Some(ref mut pm) => pm.insert(ptr as usize, layout.size(), false),
             None => {
                 let mut al = AllocList::new();
-                al.insert(ptr as usize, layout.size());
+                al.insert(ptr as usize, layout.size(), false);
                 ALLOC_INFO = Some(al);
             }
         };
@@ -91,7 +91,10 @@ unsafe impl GlobalAlloc for GCMalloc {
         if ptr.is_null() {
             // realloc is equivalent to malloc if called with a null pointer,
             // thus we must insert the new block info in the ALLOC_INFO list.
-            ALLOC_INFO.as_mut().unwrap().insert(ptr as usize, new_size);
+            ALLOC_INFO
+                .as_mut()
+                .unwrap()
+                .insert(ptr as usize, new_size, false);
         } else {
             // In all other cases, the memory block is resized and a new ptr
             // is returned. This may differ from the original ptr if the new
@@ -213,7 +216,7 @@ impl AllocList {
     /// Performs fast bump pointer insertion until the list is full, at which
     /// point, insertion is O(n) while it linearly scans the list for the next
     /// free entry.
-    pub(crate) fn insert(&mut self, ptr: usize, size: usize) {
+    pub(crate) fn insert(&mut self, ptr: usize, size: usize, gc: bool) {
         debug_assert_ne!(ptr, 0);
         debug_assert_ne!(size, 0);
 
@@ -221,7 +224,7 @@ impl AllocList {
             unsafe {
                 let next_ptr = self.start.add(self.next_free) as *mut Block;
                 if (next_ptr as usize) < self.start as usize + SIZE_ALLOC_INFO {
-                    *next_ptr = Block::Entry(core::num::NonZeroUsize::new_unchecked(ptr), size);
+                    *next_ptr = Block::Entry(core::num::NonZeroUsize::new_unchecked(ptr), size, gc);
                     self.next_free += 1;
                     return;
                 } else {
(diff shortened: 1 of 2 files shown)


src/alloc.rs
```diff
@@ -109,6 +110,53 @@ unsafe impl GlobalAlloc for GCMalloc {
...
+    unsafe fn realloc(
+        &mut self,
+        ptr: NonNull<u8>,
+        _layout: Layout,
+        new_size: usize,
+    ) -> Result<NonNull<u8>, AllocErr> {
+        let new_ptr = libc::realloc(ptr.as_ptr() as *mut libc::c_void, new_size) as *mut u8;
+
+        ALLOC_LOCK.lock();
+        // In all other cases, the memory block is resized and a new ptr
```
ltratt: I'm not quite sure how to parse "in all other cases" -- it feels like there's a missing sentence preceding this?

src/lib.rs
```diff
@@ -39,14 +39,77 @@
...
+    /// Allocate memory sufficient to `l` (i.e. correctly aligned and of at
+    /// least the required size). The returned pointer must be passed to
+    /// `Gc::from_raw`.
+    pub fn alloc_blank(l: Layout) -> *mut T {
+        let (layout, uoff) = Layout::new::<usize>().extend(l).unwrap();
+
+        unsafe {
+            let baseptr = GC_ALLOCATOR.alloc(layout).unwrap().as_ptr();
+            let objptr = baseptr.add(uoff);
+            let headerptr = objptr.sub(size_of::<usize>());
```
ltratt: It might be worth copying over (with modifications if necessary) the comment from `abgc` that explains why we do this particular dance?

ltratt:
This looks good overall but it does lack good tests -- and I suspect we can probably carry those tests forward so it might be worth spending a bit of time crafting some now. In particular, it might be worth thinking about whether there are crashing cases we want to test for (which might need something like `lang_tester`) -- it may not be sensible to include any at this point, but it's certainly worth thinking about.

src/alloc.rs
jacob-hughes: This was copied over in error. I've removed it.

src/lib.rs
jacob-hughes: Done

jacob-hughes:
I agree test coverage is a bit lacking at the moment. I'm happy to spend some time crafting them and thinking about cases where allocation may fail etc. I think it makes a bit more sense for them to come as part of the next couple PRs as they make the API to the allocator a bit clearer and we can more easily test though that. If you're okay with that?

ltratt:
OK, I can live with tests in upcoming PRs. I'm not sure that you've pushed a commit with the minor other fixes I suggested though?

jacob-hughes:
Ok sounds good. Can I push these minor comment changes as part of the squash?

ltratt:
If you do `git reset --soft 2a2c47974e68fa5f654b4`, does that then allow you to easily get you your changes as a single commit?

jacob-hughes:
Addressed comments

src/lib.rs
```diff
@@ -38,10 +38,90 @@
...
+        // sizeof::<usize>()` gives a valid alignment for a `usize`. There are
+        // only two cases we need to consider here:
+        //   1) `object`'s alignment is smaller than or equal to `usize`. If so,
+        //      no padding will be added, at which point by definition `uoff -
+        //      sizeof::<usize>()` will be exactly equivalent to the start point
+        //      of the layout.
+        //   2) `object`'s alignment is bigger than `usize`. Since alignment
+        //      must be a power of two, that means that we must by definition be
+        //      adding at least one exact multiple of `usize` bytes of padding.
+        //      The assert below is thus paranoia writ large: it could only
```
ltratt: Since you (perfectly reasonably) don't have this `assert`, I think you can just remove this sentence.

jacob-hughes: Okay, done.

src/lib.rs
```diff
@@ -86,18 +86,10 @@ impl<T> Gc<T> {
...
-        // sizeof::<usize>()` gives a valid alignment for a `usize`. There are
-        // only two cases we need to consider here:
-        //   1) `object`'s alignment is smaller than or equal to `usize`. If so,
-        //      no padding will be added, at which point by definition `uoff -
-        //      sizeof::<usize>()` will be exactly equivalent to the start point
-        //      of the layout.
-        //   2) `object`'s alignment is bigger than `usize`. Since alignment
-        //      must be a power of two, that means that we must by definition be
-        //      adding at least one exact multiple of `usize` bytes of padding.
-        //      The assert below is thus paranoia writ large: it could only
```
ltratt: I only meant you should delete the sentence "the assert below". Case (2) is still important for you, but you just don't have the `assert`!

jacob-hughes: Oh I see. Sorry, fixed that.

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #5, opened 2019-10-10
Add lang_test framework for testing full programs

jacob-hughes:
With conservative scanning, a proper testing framework is necessary as we want to run programs from start to finish. This PR sets up `lang_tester` for this purpose. There aren't many tests here yet because they won't make sense without the stack scanning code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,13 @@ edition = "2018"
 
 [dependencies]
 libc = "*"
+
+[dev-dependencies]
+lang_tester = "0.3"
+tempdir = "0.3"
+
+[[test]]
+name = "gc_tests"
+path = "gc_tests/run_tests.rs"
+harness = false
+
--- a/gc_tests/run_tests.rs
+++ b/gc_tests/run_tests.rs
@@ -1,0 +1,85 @@
+// Copyright (c) 2019 King's College London created by the Software Development
+// Team <http://soft-dev.org/>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0>, or the MIT license <LICENSE-MIT
+// or http://opensource.org/licenses/MIT>, or the UPL-1.0 license
+// <http://opensource.org/licenses/UPL> at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use std::{env, path::PathBuf, process::Command};
+
+use lang_tester::LangTester;
+use tempdir::TempDir;
+
+static CRATE_NAME: &'static str = "gcmalloc";
+
+fn main() {
+    // We grab the rlibs from `target/<debug | release>/` but in order
+    // for them to exist here, they must have been moved from `deps/`.
+    // Simply running `cargo test` will not do this, instead, we must
+    // ensure that `cargo build` has been run before running the tests.
+    Command::new("cargo")
+        .args(&["build"])
+        .output()
+        .expect("Failed to build libs");
+
+    let tempdir = TempDir::new("gc_tester").unwrap();
+    LangTester::new()
+        .test_dir("gc_tests/tests")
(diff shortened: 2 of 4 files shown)


jacob-hughes:
We now run `cargo build` as part of the `lang_tester` setup which should ensure that the rlibs are built and moved into the correct place.

This PR is now ready for review

Cargo.toml
```diff
@@ -6,3 +6,13 @@ edition = "2018"
 
 [dependencies]
 libc = "*"
+
+[dev-dependencies]
+lang_tester = "0.3"
+tempdir = "*"
```
ltratt: Is there a reason we're not using a specific version of this library?

gc_tests/run_tests.rs
```diff
@@ -0,0 +1,88 @@
...
+fn main() {
+    let tempdir = TempDir::new("gc_tester").unwrap();
+    LangTester::new()
+        .test_dir("gc_tests/tests")
+        .test_file_filter(|p| p.extension().unwrap().to_str().unwrap() == "rs")
+        .test_extract(|s| {
+            Some(
+                s.lines()
+                    .skip_while(|l| !l.starts_with("//"))
+                    .take_while(|l| l.starts_with("//"))
```
ltratt: I think you only need the `skip_while` or the `take_while`, but not both (since they're inverses of each other).

gc_tests/run_tests.rs
```diff
@@ -0,0 +1,88 @@
...
+                    .collect::<Vec<_>>()
+                    .join("\n"),
+            )
+        })
+        .test_cmds(move |p| {
+            // We grab the rlibs from `target/<debug | release>/` but in order
+            // for them to exist here, they must have been moved from `deps/`.
+            // Simply running `cargo test` will not do this, instead, we must
+            // ensure that `cargo build` has been run before running the tests.
+            let mut build_libs = Command::new("cargo");
```
ltratt: This will run `cargo build` for every lang test: can't it be run just once, up front?

gc_tests/tests/alloc_info_oom.rs
```diff
@@ -0,0 +1,13 @@
+// Run-time:
+//  status: error
```
ltratt: It might not be a bad idea to test for specific `stderr` output here? Even something simple like:

```
stderr: ...out of memory...
```

just so that other types of errors don't get confused for the one you're intending to test.

Cargo.toml
jacob-hughes: Yes, laziness ;) I've set this to `0.3.7`

ltratt: We should generally use just the major and minors so `0.3` would be better (cargo treats the patch level as a minimum requirement, so it's almost never worth specifying IMHO).

gc_tests/tests/alloc_info_oom.rs
jacob-hughes: Good point. I'll add this.

Cargo.toml
jacob-hughes: I see

jacob-hughes:
Addressed @ltratt's comments.

src/alloc.rs
```diff
@@ -377,8 +377,10 @@ pub struct PtrInfo {
     pub gc: bool,
 }
 
-fn abort() {
-    unsafe { core::intrinsics::abort() };
+fn abort(msg: &str) {
```
ltratt: It might not be a good idea to have a function which matches a well-known builtin (`abort`) but with different semantics (this `abort` cleans up after itself). Maybe name it something like `alloc_exit`?

jacob-hughes: Good idea

jacob-hughes:
Addressed all comments

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #6, opened 2019-10-22
Implement root scanning for garbage collected objects

jacob-hughes:
_[It may make reviewing a lot easier to look at this PR in chronological commit order; the diff for  f4f83d6 is quite large, but it simply hides some things which were public when they shouldn't have been and creates a single API to the alloc metadata which can be used by the GC.]_

Users can trigger a single collection through a public `collect()` function, which, at present, simply marks objects reachable directly from the root-set. Reclamation of unreachable GC objects is not yet
implemented.

We define the root-set as any reference live on the stack during a collection or values inside callee-save registers. Since register spilling is inherently platform specific, we require assembly stubs containing the spill code for each architecture. At present, we only implement X86_64.

A secondary debug API exists for helping query information about a collection for tests etc.

The collector is **not** in anyway thread-safe (yet).

There's a fair bit of assembly magic here, so I think it might be wise to cc @vext01 on this one too.

### DIFF
--- a/gc_tests/tests/on_stack_root_marking.rs
+++ b/gc_tests/tests/on_stack_root_marking.rs
@@ -1,0 +1,22 @@
+// Run-time:
+//  status: success
+
+extern crate gcmalloc;
+
+use gcmalloc::Gc;
+use gcmalloc::{collect, Debug };
+use gcmalloc::gc::DebugFlags;
+
+fn foo() {
+    let y = Gc::new(456 as usize);
+    collect();
+    assert!(Debug::is_black(y));
+}
+
+fn main() {
+    gcmalloc::init(DebugFlags::new().mark_only());
+
+    let x = Gc::new(123 as usize);
+    foo(); // triggers a collection
+    assert!(Debug::is_black(x));
+}
--- a/src/SpillRegisters_X64.S
+++ b/src/SpillRegisters_X64.S
@@ -1,0 +1,30 @@
+.text
+
+.extern scan_stack
+.globl spill_registers
+
+spill_registers:
+    // Push all callee-save registers to get them on the stack for conservative
+    // stack scanning. We maintain 16-byte alignment at calls (Sys V ABI). There
+    // is an 8-byte return address on the stack and we push 56 bytes which
+    // maintains 16-byte stack alignment at the call.
+    //
+    // We don't need to push caller-save registers, as they will have already
+    // been spilled if the caller requires them to outlive the call.
+    push $0
+    push %rbx
+    push %rbp
+    push %r12
+    push %r13
+    push %r14
+    push %r15
+    // %rdi already contains the 'self' reference to the collector. We move the
+    // function pointer in %rsi to %r8 (a scratch register) so that the stack
+    // pointer can be placed in %rdi as the first arg slot.
+    mov %rsi, %r8
+    mov %rsp, %rsi
+    call *%r8
+    // Pop all the callee-save registers (ret will pop the return addr)
+    add $56, %rsp
+    ret
+
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -43,7 +43,7 @@ use std::{
 
 static SIZE_ALLOC_INFO: usize = (1024 * 1024) * 2; // 2MiB
 
-pub(crate) static mut ALLOC_INFO: Option<AllocList> = None;
+static mut ALLOC_INFO: Option<AllocList> = None;
 
 /// A spinlock for the global ALLOC_INFO list.
 ///
@@ -116,16 +116,6 @@ unsafe impl Alloc for GCMalloc {
     unsafe fn alloc(&mut self, layout: Layout) -> Result<NonNull<u8>, AllocErr> {
         let ptr = libc::malloc(layout.size() as libc::size_t) as *mut u8;
 
-        ALLOC_LOCK.lock();
-        match ALLOC_INFO {
-            Some(ref mut pm) => pm.insert(ptr as usize, layout.size(), true),
-            None => {
-                let mut al = AllocList::new();
-                al.insert(ptr as usize, layout.size(), true);
-                ALLOC_INFO = Some(al);
-            }
-        };
-        ALLOC_LOCK.unlock();
         Ok(NonNull::new_unchecked(ptr))
     }
 
@@ -154,20 +144,20 @@ unsafe impl Alloc for GCMalloc {
     }
 }
 
-pub(crate) struct AllocLock(AtomicBool);
+struct AllocLock(AtomicBool);
 
 impl AllocLock {
-    pub(crate) const fn new() -> AllocLock {
+    const fn new() -> AllocLock {
         Self(AtomicBool::new(false))
     }
 
-    pub(crate) fn lock(&self) {
+    fn lock(&self) {
         while !self.0.compare_and_swap(false, true, Ordering::AcqRel) {
             // Spin
         }
     }
 
-    pub(crate) fn unlock(&self) {
+    fn unlock(&self) {
         self.0.store(false, Ordering::Release);
     }
 }
@@ -182,7 +172,7 @@ impl AllocLock {
 /// conservative GC to know which additional machine words must be scanned.
 ///
 /// TODO: Grow the AllocList if it exceeds its size.
-pub(crate) struct AllocList {
+struct AllocList {
     start: *const Block,
     next_free: usize,
     can_bump: bool,
@@ -235,7 +225,7 @@ impl<'a> Iterator for AllocListIter<'a> {
 }
 
 impl AllocList {
-    pub(crate) fn new() -> AllocList {
+    fn new() -> AllocList {
         let raw = unsafe { libc::malloc(SIZE_ALLOC_INFO as libc::size_t) } as *const Block;
         AllocList {
             start: raw,
@@ -261,7 +251,7 @@ impl AllocList {
     /// Performs fast bump pointer insertion until the list is full, at which
     /// point, insertion is O(n) while it linearly scans the list for the next
     /// free entry.
-    pub(crate) fn insert(&mut self, ptr: usize, size: usize, gc: bool) {
+    fn insert(&mut self, ptr: usize, size: usize, gc: bool) {
         debug_assert_ne!(ptr, 0);
         debug_assert_ne!(size, 0);
 
@@ -297,7 +287,7 @@ impl AllocList {
 
     /// Remove ptr information associated with a base pointer (perfomed on a
     /// dealloc). This must not be called with an inner pointer.
-    pub(crate) fn remove(&mut self, ptr: usize) {
+    fn remove(&mut self, ptr: usize) {
         for block in self.iter_mut() {
             if let Block::Entry(base_ptr, ..) = block {
                 if base_ptr.get() == ptr {
@@ -320,7 +310,7 @@ impl AllocList {
     /// In conservative GC, this guarantee is implicit as this method is only
     /// ever called during a stop-the-world stack scanning phase where we can be
     /// certain no mutator threads are running.
-    pub(crate) fn find_base(&self, ptr: usize) -> Option<PtrInfo> {
+    fn find_base(&self, ptr: usize) -> Option<PtrInfo> {
         self.iter().find_map(|x| {
             if let Block::Entry(base_ptr, size, gc) = *x {
                 if ptr >= base_ptr.get() && ptr < (base_ptr.get() + size as usize) {
@@ -337,7 +327,7 @@ impl AllocList {
 
     /// Updates the size associated with a base pointer (perfomed on a realloc).
     /// This must not be called with an inner pointer.
-    pub(crate) fn update(&mut self, ptr: usize, new_ptr: usize, size: usize) {
+    fn update(&mut self, ptr: usize, new_ptr: usize, size: usize) {
         for block in self.iter_mut() {
             if let Block::Entry(base_ptr, _size, gc) = *block {
                 if ptr == base_ptr.get() {
@@ -380,6 +370,58 @@ pub struct PtrInfo {
 fn exit_alloc(msg: &str) {
     unsafe { libc::write(2, msg.as_ptr() as *const libc::c_void, msg.len()) };
     std::process::exit(1);
+}
+
+// -----------------------------------------------------------------------------
+//        APIs for querying the block information about each allocation
+// -----------------------------------------------------------------------------
+
+pub struct AllocMetadata;
+
+impl AllocMetadata {
+    /// Insert metadata about an allocation block. The GC scans for reachable
+    /// objects conservatively. It uses the metadata stored against each
+    /// allocation to determine:
+    /// 1) the size of an allocation block
+    /// 2) whether an allocation block's memory management is the responsibility
+    ///    of the GC
+    /// 3) whether an arbitrary pointer-like word points within an
+    ///    allocation block
+    pub(crate) fn insert(ptr: usize, size: usize, gc: bool) {
+        ALLOC_LOCK.lock();
+        unsafe {
+            match ALLOC_INFO {
+                Some(ref mut pm) => pm.insert(ptr as usize, size, true),
+                None => {
+                    let mut al = AllocList::new();
+                    al.insert(ptr as usize, size, gc);
+                    ALLOC_INFO = Some(al);
+                }
+            };
+        }
+        ALLOC_LOCK.unlock();
+    }
+    /// Returns metadata about an allocation block when given an arbitrary
+    /// pointer to the start of the block or an offset within it.
+    pub(crate) fn find(ptr: usize) -> Option<PtrInfo> {
+        ALLOC_LOCK.lock();
+        let ptr_info = unsafe { ALLOC_INFO.as_ref().unwrap().find_base(ptr) };
+        ALLOC_LOCK.unlock();
+        ptr_info
+    }
+
+    /// Removes the metadata associated with an allocation.
+    pub(crate) fn remove(ptr: usize) {
+        ALLOC_LOCK.lock();
+
+        debug_assert!(|| -> bool {
+            let ptr_info = unsafe { ALLOC_INFO.as_ref().unwrap().find_base(ptr).unwrap() };
+            ptr_info.gc
+        }());
+
+        unsafe { ALLOC_INFO.as_mut().unwrap().remove(ptr) };
+        ALLOC_LOCK.unlock();
+    }
 }
 
 #[cfg(test)]
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,0 +1,192 @@
+// Copyright (c) 2019 King's College London created by the Software Development
+// Team <http://soft-dev.org/>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0>, or the MIT license <LICENSE-MIT
+// or http://opensource.org/licenses/MIT>, or the UPL-1.0 license
+// <http://opensource.org/licenses/UPL> at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+use crate::{
+    alloc::{AllocMetadata, PtrInfo},
+    Gc,
+};
+use std::sync::Mutex;
+
+static WORD_SIZE: usize = 8; // Bytes
+
+type StackAddress = usize;
+type StackValue = usize;
+
+type StackScanCallback = extern "sysv64" fn(&mut Collector, StackAddress);
+#[link(name = "SpillRegisters", kind = "static")]
+extern "sysv64" {
+    // Pass a type-punned pointer to the collector and move it to the asm spill
+    // code. This is so it can be passed straight back as the implicit `self`
+    // address in the callback.
+    #[allow(improper_ctypes)]
+    fn spill_registers(collector: *mut u8, callback: StackScanCallback);
+}
+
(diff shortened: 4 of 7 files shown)


src/SpillRegisters_X64.S
```diff
@@ -0,0 +1,27 @@
+.text
+
+.extern scan_stack
+.globl spill_registers
+
+spill_registers:
+    // Push all callee-saves registers to get them on the stack for conservative
```
vext01: Typo: "callee-save"

src/SpillRegisters_X64.S
```diff
@@ -0,0 +1,27 @@
...
+    // stack scanning. We maintain 16-byte alignment at calls (Sys V ABI). There
+    // is an 8-byte return address on the stack and we push 56 bytes which
+    // maintains 16-byte stack alignment at the call.
+    push $0
```
vext01: What's `$0` again?

src/SpillRegisters_X64.S
vext01: Did you want to document why it's only the callee save registers that we care about?

src/SpillRegisters_X64.S
jacob-hughes: GAS syntax requires all constants to be prefixed with `$`. We're simply pushing padding to the stack to maintain 16-byte alignment.

[As an aside, I ran into a nasty gotcha with this one. I'd written `push 0` before, which actually pushes the contents of memory address 0, which, of course, segfaults as it tries to dereference it. :man_facepalming:  ]

src/SpillRegisters_X64.S
jacob-hughes: Sure, done.

src/SpillRegisters_X64.S
vext01: Ah yes. Good old GNU syntax.

src/SpillRegisters_X64.S
vext01: LGTM

src/alloc.rs
```diff
@@ -382,6 +372,58 @@ fn exit_alloc(msg: &str) {
...
+                None => {
+                    let mut al = AllocList::new();
+                    al.insert(ptr as usize, size, gc);
+                    ALLOC_INFO = Some(al);
+                }
+            };
+        }
+        ALLOC_LOCK.unlock();
+    }
+    /// Returns metadata about an allocation block when given an arbitrary
```
ltratt: Since this function returns `Option`, I assume you can pass *any* pointer in here? The doc string suggests that only valid pointers can be passed in but I don't think that's correct.

jacob-hughes: Yes, I see what you mean by the wording of the docstring. I'll reword it to make it clearer that you _can_ pass any pointer. But it will only return pointer info in those circumstances.

src/alloc.rs
```diff
@@ -382,6 +372,58 @@ fn exit_alloc(msg: &str) {
...
+        ALLOC_LOCK.unlock();
+        ptr_info
+    }
+
+    /// Removes the metadata associated with an allocation.
+    pub(crate) fn remove(ptr: usize) {
+        ALLOC_LOCK.lock();
+        let ptr_info = unsafe { ALLOC_INFO.as_ref().unwrap().find_base(ptr).unwrap() };
+
+        debug_assert!(ptr_info.gc); // FIXME: Move this expensive function into the assertion
```
ltratt: I don't understand the assertion: the `assert` isn't a function call. You can, from memory, do things like:

```rust
#[cfg(debug)]
{
    let ptr_info = ...;
    assert!(ptr_info.gc);
}
```

if that's what you wanted?

src/lib.rs
```diff
@@ -101,12 +124,38 @@ impl<T> Gc<T> {
...
     }
 }
 
+impl GcHeader {
+    pub(crate) fn new() -> Self {
+        Self(0)
+    }
+
+    pub(crate) fn mark_bit(&self) -> bool {
+        ((self.0 >> 0) & 1) == 1
```
ltratt: Isn't `n >> 0` simply `n`?

src/alloc.rs
jacob-hughes: Sorry, that comment shouldn't have ended up in the PR and was a reminder for me to do exactly what you suggested.

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use crate::{
+    alloc::{AllocMetadata, PtrInfo},
+    Gc,
+};
+use std::sync::Mutex;
+
+static WORD_SIZE: usize = 8; // Bytes
```
ltratt: I suggest you guard this with `#[cfg(target_pointer_width = "64")]` so that if anyone tries compiling this on <>64 bit machine they'll get a static error.

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+
+type StackAddress = usize;
+type StackValue = usize;
+
+type StackScanCallback = extern "sysv64" fn(&mut Collector, StackAddress);
+#[link(name = "SpillRegisters", kind = "static")]
+extern "sysv64" {
+    // Rust (correctly) warns us that `Collector` has an unspecified layout and
+    // that it is UB to pass this to C code via FFI. However, our register spill
+    // code does not dereferences this, nor does it ever move the `Collector`
```
ltratt: "dereference"s -> "dereference".

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+static WORD_SIZE: usize = 8; // Bytes
+
+type StackAddress = usize;
+type StackValue = usize;
+
+type StackScanCallback = extern "sysv64" fn(&mut Collector, StackAddress);
+#[link(name = "SpillRegisters", kind = "static")]
+extern "sysv64" {
+    // Rust (correctly) warns us that `Collector` has an unspecified layout and
+    // that it is UB to pass this to C code via FFI. However, our register spill
```
ltratt: Why don't we pass it as, say, a `const *u8` which will probably get rid of the warning?

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+pub struct DebugFlags {
+    mark_only: bool,
+}
+
+impl DebugFlags {
+    pub fn new() -> Self {
+        Self { mark_only: false }
+    }
+
+    pub fn none() -> Self {
```
ltratt: What's the point of this function?

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    pub(crate) fn collect(&mut self) {
+        // First check that no call to collect is active
+        {
+            let mut cstate = self.state.lock().unwrap();
+            match *cstate {
+                CollectorState::Ready => *cstate = CollectorState::Marking,
+                _ => {
+                    // Bail out if we reach this point as the collector has already
+                    // been invoked -- and is running -- on another thread.
```
ltratt: Simply saying "The collector is running in another thread" is probably good enough here?

src/lib.rs
```diff
@@ -156,12 +165,53 @@ impl<T> Deref for Gc<T> {
...
 
+pub fn init(flags: gc::DebugFlags) {
+    unsafe { COLLECTOR = Some(Collector::new(flags)) };
+}
+
+pub fn collect() {
+    // FIXME: Is this UB? Theoretically this pub `collect` fn can be called in
+    // parallel from multiple threads, allowing us potentially multiple mutable
+    // references to the underlying `COLLECTOR` struct. However, the inner
+    // `collect` method guards against multiple collection invocations via a
```
ltratt: Why would this be UB? Merely having a reference to a struct in multiple threads cannot lead to UB (unless the underlying memory moves or similar!): it's reading and writing that one has to be careful about.

src/lib.rs
```diff
@@ -156,12 +165,53 @@ impl<T> Deref for Gc<T> {
...
+}
+
+pub struct Debug;
+
+impl Debug {
+    pub fn is_black<T>(gc: Gc<T>) -> bool {
+        let collector = unsafe { COLLECTOR.as_ref().unwrap() };
+        let cstate = *collector.state.lock().unwrap();
+
+        // Checking an object colour only makes sense immediately after marking has
```
ltratt: "object colour" -> "object's colour"

src/lib.rs
```diff
@@ -156,12 +165,53 @@ impl<T> Deref for Gc<T> {
...
+        // taken place. After a full collection has happened, marking results are
+        // stale and the object graph must be re-marked in order for this query
+        // to be meaningful.
+        assert_eq!(cstate, gc::CollectorState::FinishedMarking);
+        return collector.colour(unsafe { Gc::from_raw(gc.objptr as *const i8) })
+            == gc::Colour::Black;
+    }
+
+    /// This should be used with caution. For example, if writing a test using
```
ltratt: This doc string feels a bit confusing. Why not just say something like "If true, this object was not marked for collection in the last sweep"? [BTW, in a sense, your whole warning also applies to `is_black`, since the two functions are simply inverses.]

jacob-hughes: I've reworded this.

I think the warning isn't as relevant to `is_black` as the conservative GC will tend to over-approximate - if I assert an object should be black after a collection and it's not, that's a bug! 

On the other hand, I can't be sure that an object didn't get marked, as for some non-deterministic reason, residual pointers may have been left lingering on the stack or within registers.

src/lib.rs
jacob-hughes: Fixed

src/lib.rs
jacob-hughes: Makes sense. I've removed the comment.

src/gc.rs
jacob-hughes: Fixed

src/gc.rs
jacob-hughes: Premature over-engineering. It's gone

src/gc.rs
jacob-hughes: Okay

src/gc.rs
jacob-hughes: Fixed

src/gc.rs
jacob-hughes: I've decided to guard the entire library, as there are other places where this assumption is made too.

src/gc.rs
```diff
@@ -0,0 +1,229 @@
+// Copyright (c) 2019 King's College London
```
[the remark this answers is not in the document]
jacob-hughes: Okay done. For future reference, should the main `lib.rs` / `mod.rs` file only contain the full license?

src/lib.rs
jacob-hughes: Yes. Fixed

src/lib.rs
ltratt: I think none of this is a problem if you avoid talking about e.g "liveness" (so, for example, it would definitely have been incorrect to say "this guarantees the object was live at the last collection). That said, what can anyone do with such a guarantee? I think the answer is "nothing": by definition, the instant the collection is finished, this information is stale and can't be relied on. So I think you might be worrying unduly :)

src/gc.rs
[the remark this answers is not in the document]
ltratt: What we've gone for, mostly, is a `LICENSE` file and then a short header at the top of each file e.g. see `Vob` https://github.com/softdevteam/vob/blob/master/src/lib.rs

src/alloc.rs
```diff
@@ -413,9 +413,11 @@ impl AllocMetadata {
     /// Removes the metadata associated with an allocation.
     pub(crate) fn remove(ptr: usize) {
         ALLOC_LOCK.lock();
-        let ptr_info = unsafe { ALLOC_INFO.as_ref().unwrap().find_base(ptr).unwrap() };
 
-        debug_assert!(ptr_info.gc); // FIXME: Move this expensive function into the assertion
+        debug_assert!(|| -> bool {
```
ltratt: Ha, that is not a way of doing it that had occurred to me!

src/gc.rs
```diff
@@ -21,15 +21,10 @@ type StackValue = usize;
...
-    // instance by value. It simply moves the address to the first argument
-    // register to be used as the implicit `self` address in the callback. It is
-    // therefore safe to do this without specifying a C style layout for the
-    // struct.
-    #[allow(improper_ctypes)]
-    fn spill_registers(collector: &mut Collector, callback: StackScanCallback);
+    // Pass a type-punned pointer to the collector and move it to the asm spill
+    // code. This is so it can be passed straight back as the implicit `self`
+    // address in the callback.
+    fn spill_registers(collector: *mut i8, callback: StackScanCallback);
```
ltratt: I think it's more conventional to use `*mut u8`, but I might be wrong.

src/lib.rs
jacob-hughes: Sure. I suppose the main point that I am trying to make it clear is that it's safe to use `is_black` in a test, but not `is_white`. It's always safe to over-approximate, and it doesn't really matter if a white object was coloured black as a false positive. However, we can't run a test collection and assert that objects which we didn't expect to be coloured black remain white.

src/gc.rs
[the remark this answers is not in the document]
jacob-hughes: Okay, so I should really replace the license with the shorter form in the other files too?

src/gc.rs
jacob-hughes: Sure, I think I probably meant to write `u8` anyway.

ltratt:
@jacob-hughes I assume you meant to remind me to rereview this?

src/gc.rs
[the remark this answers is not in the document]
ltratt: Yeah. You can do it in another PR if that makes more sense from your perspective.

src/lib.rs
jacob-hughes: > Unless, that is, one explicitly says "you can not safely use !o.is_black()"?

Exactly `!is_black()` is not something you should ever test for. Perhaps I should make that clearer in the doc comments. It's only right to test that a collector reached reachable objects in the last collection.

jacob-hughes:
Yes sorry

src/alloc.rs
```diff
@@ -116,16 +116,6 @@ unsafe impl Alloc for GCMalloc {
     unsafe fn alloc(&mut self, layout: Layout) -> Result<NonNull<u8>, AllocErr> {
         let ptr = libc::malloc(layout.size() as libc::size_t) as *mut u8;
 
-        ALLOC_LOCK.lock();
```
vext01: So how comes you no longer record the allocations here?

src/lib.rs
```diff
@@ -98,6 +98,8 @@ impl<T> Gc<T> {
         unsafe {
             let baseptr = GC_ALLOCATOR.alloc(layout).unwrap().as_ptr();
             let objptr = baseptr.add(uoff);
+
+            AllocMetadata::insert(objptr as usize, layout.size(), true);
```
vext01: I see, so you do the insertion here...

vext01:
Typo in the first commit message "...to get interact with allocation block metadata". Get or interact?

src/lib.rs
jacob-hughes: Yes, and it's subtly different. We insert the `objptr` for GC objects, not the `baseptr`. This means that theoretically, if there existed only a pointer on the stack to a GC object's `baseptr`, it would not be considered reachable and would be reclaimed.

src/lib.rs
```diff
@@ -101,12 +124,38 @@ impl<T> Gc<T> {
...
+        Self(0)
+    }
+
+    pub(crate) fn mark_bit(&self) -> bool {
+        ((self.0 >> 0) & 1) == 1
+    }
+
+    pub(crate) fn set_mark_bit(&mut self, mark: bool) {
+        if mark {
+            self.0 |= 1 << 0
```
vext01: More zero shifts here.

src/lib.rs
```diff
@@ -101,12 +124,38 @@ impl<T> Gc<T> {
...
+        } else {
+            self.0 &= !(1 << 0)
```
vext01: I think it was naughty of Rust to use `!` for bitwise compliment on integers instead of `~`.

jacob-hughes: I agree

src/lib.rs
```diff
@@ -101,12 +124,38 @@ impl<T> Gc<T> {
...
+        }
+    }
+}
+
+impl<T> Deref for Gc<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        unsafe { &*(self.objptr as *const T) }
```
vext01: What's unsafe here? `&*`?

gc_tests/tests/on_stack_root_marking.rs
```diff
@@ -0,0 +1,24 @@
...
+use gcmalloc::{collect, Debug };
+use gcmalloc::gc::DebugFlags;
+
+fn foo() {
+    let y = Gc::new(456 as usize);
+    collect();
+    assert!(Debug::is_black(y));
+}
+
+fn main() {
```
vext01: quite a lot of whitespace here.

src/lib.rs
jacob-hughes: Yes, it's de-referencing a raw pointer.

src/gc.rs
vext01: Laurie an I agreed to exhaustively enumerate (without `_`) in cases where it is practical, as it forces us to consider behaviours in the case that we add another variant.

src/gc.rs
```diff
@@ -0,0 +1,229 @@
...
+                    break;
+                }
+
+                self.mark(obj, Colour::Black);
+            }
+        }
+    }
+
+    #[no_mangle]
+    extern "sysv64" fn scan_stack(&mut self, rsp: StackAddress) {
```
vext01: Why are we placing ABI constraints here? Does C code call this?

src/gc.rs
jacob-hughes: That makes sense,  but in this instance, collection can never happen when the collector is in any state other than `CollectorState::Ready`, even if new variants are added.

src/lib.rs
vext01: As long as there are no aliasing problems... I have no idea.

src/gc.rs
jacob-hughes: Because this function is called from the `SpillRegisters_x86_64` stub which sets up its arguments according to SYS V.

src/lib.rs
vext01: Did you try without? It may not actually deref.

src/lib.rs
ltratt: OK, if that's the constraint then stating it explicitly is a good idea. The explanation doesn't have to be long, so long as it's clear.

src/lib.rs
jacob-hughes: In Rust you have to compose both operators if you want to get a reference from a raw pointer. It doesn't deref now, but the references are always considered dereference-able, so if you construct one from a raw pointer you better be sure that it's safe to de-reference it in the future.

vext01: Right, but that didn't answer my question. Did you try it without?

(You are probably right)

jacob-hughes: I'm not sure what you mean? `&(self.objptr as *const T)` is invalid Rust code.

ltratt: Turning a raw pointer into a Rust reference is always unsafe.

src/gc.rs
ltratt: Yeah, in this situation `_` did feel like it was reasonable to me (I explicitly thought about it!).

src/lib.rs
vext01: > Turning a raw pointer into a Rust reference is always unsafe.

ok

jacob-hughes: @vext01 Sorry, I understand now that you meant try and remove the unsafe block, not the `*`.

src/gc.rs
vext01: It's up to you two. Seemed like a good candidate for the rule to me (few variants, not inconceivable that a new GC state be added later).

src/lib.rs
jacob-hughes: I've decided to get rid of `is_white` altogether to avoid confusion, and warn on using the inverse of `is_black`.

gc_tests/tests/on_stack_root_marking.rs
jacob-hughes: Fixed

src/lib.rs
jacob-hughes: Fixed

vext01:
LGTM

src/lib.rs
```diff
@@ -179,6 +179,14 @@ pub fn collect() {
 pub struct Debug;
 
 impl Debug {
+    /// Checks that the object was marked as reachable in the last collection.
```
ltratt: It's a bit difficult to know what true/false means here. How about saying `Returns `true` if this object was reachable in the last collection`?

jacob-hughes: Sure

jacob-hughes:
Ready for re-review

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
Is " Add single API to interact with allocation block metadata " supposed to be "Add *simple* API to ..."?

In https://github.com/softdevteam/gcmalloc/pull/6/commits/0a16bd587fa123fc7dbd70d04adffea92afa270e "not in anyway" -> "not in any way".

Feel free to force push updates.

jacob-hughes:
Yes, not quite sure how I missed that one. Fixed

PR #7, opened 2019-10-24
Add triple license

jacob-hughes:
And update the file headers to use the shorter format


jacob-hughes:
I'll take your word for it. I'll do what `Vob` does because it means that the project shows up in the Github GUI as being licensed.

jacob-hughes:
Fixed

ltratt:
Please squash.

jacob-hughes:
Squashed. I've updated the PR title too so hopefully your merge commit matches

PR #8, opened 2019-10-25
Use rerun_except to stop lang_tests forcing unnecessary rebuilds.

ltratt:
Changing a lang test in `gc_tests/tests` currently triggers a full rebuild of `gcmalloc`, despite that being unnecessary. Using `rerun_except` stops this happening.


jacob-hughes:
Oops! I never saw this one in the PR channel. 

Thanks, this looks good!

PR #9, opened 2019-10-28
Trace through allocation blocks during collection

jacob-hughes:
After stack scanning, where the root-set is added to the marking worklist, the collector traces through each heap allocation looking for further pointers. This completes the necessary steps for the mark phase.

The other important change in this PR is that we make `Gc<T>` copyable for all values of `T`. This seriously improves its ergonomics (as seen in the various new `gc_tests`). It means that `Gc<T>` is now a real viable alternative for writing cyclic data structures. Unlike `Rc<T>` and friends, we can make this trivially copyable because we don't need to implement `Drop`. The collector cleans up for us at some non-deterministic point in the future. That's not to say we won't need some alternative form of finalization in the future; but it won't be `Drop` because it's not bound by Rust's RAII guarantees.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -13,12 +13,13 @@ use crate::{
 };
 use std::sync::Mutex;
 
-static WORD_SIZE: usize = 8; // Bytes
+static WORD_SIZE: usize = std::mem::size_of::<usize>(); // Bytes
 
-type StackAddress = usize;
-type StackValue = usize;
+type Address = usize;
 
-type StackScanCallback = extern "sysv64" fn(&mut Collector, StackAddress);
+type Word = usize;
+
+type StackScanCallback = extern "sysv64" fn(&mut Collector, Address);
 #[link(name = "SpillRegisters", kind = "static")]
 extern "sysv64" {
     // Pass a type-punned pointer to the collector and move it to the asm spill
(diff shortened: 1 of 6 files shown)


src/gc.rs
```diff
@@ -133,20 +134,28 @@ impl Collector {
...
                 if self.colour(obj) == Colour::Black {
-                    break;
+                    continue;
                 }
-
                 self.mark(obj, Colour::Black);
             }
+
+            // Check each word in the allocation block for pointers.
+            for addr in (ptr..(ptr + size)).step_by(WORD_SIZE) {
```
ltratt: Are the inner brackets `(ptr + size)` necessary? I guess not, but maybe worth checking with `rust fmt`?

ltratt:
This PR looks fine, so I assume it's merge as-is?

However, the next thing we do needs to be (in order):

1) Remove the warnings (unused methods). Going forward, we shouldn't ever be committing code with warnings.
2) Run valgrind against gcmalloc and fix what look like bugs (e.g. there are invalid reads and writes; gcmalloc segfaults on OpenBSD, probably because of these problems). Some of these are probably minor, but some might be major, and if we let them fester for too long, we might get into trouble!

src/gc.rs
jacob-hughes: They are not necessary no. `rust fmt` doesn't care either way. Personally I prefer them as I think it makes it more readable, but I'm not that fussed either way if you think I should remove them.

jacob-hughes: Sure. They're gone now

jacob-hughes:
Ready to squash when you need.

> However, the next thing we do needs to be (in order) ... 

PR upcoming :)

ltratt:
Please squash and rebase.

jacob-hughes:
Squashed.

PR #12, opened 2019-11-04
Refactor alloc metadata calls.

jacob-hughes:
When tracking down some bugs as part of the sanitizer support PR, I had refactored the allocation metadata into a single API to see if that was responsible for some of my concurrency bugs. Alas, it was not, but it simplifies the code anyway and results in some deletion so it's worth checking in.

The unit tests have also been refactored to use the global allocator API too, which provides better coverage when running the tests with TSAN. This also ensures that `AllocMetadata::remove` is tested correctly and no longer dead code.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -37,48 +37,22 @@ pub struct AllocWithInfo;
 unsafe impl GlobalAlloc for AllocWithInfo {
     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
         let ptr = libc::malloc(layout.size() as libc::size_t) as *mut u8;
-
-        ALLOC_LOCK.lock();
-        match ALLOC_INFO {
-            Some(ref mut pm) => pm.insert(ptr as usize, layout.size(), false),
-            None => {
-                let mut al = AllocList::new();
-                al.insert(ptr as usize, layout.size(), false);
-                ALLOC_INFO = Some(al);
-            }
-        };
-        ALLOC_LOCK.unlock();
+        AllocMetadata::insert(ptr as usize, layout.size(), false);
         ptr
     }
 
     unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
         libc::free(ptr as *mut libc::c_void);
-        ALLOC_LOCK.lock();
-        ALLOC_INFO.as_mut().unwrap().remove(ptr as usize);
-        ALLOC_LOCK.unlock();
+        AllocMetadata::remove(ptr as usize);
     }
 
     unsafe fn realloc(&self, ptr: *mut u8, _layout: Layout, new_size: usize) -> *mut u8 {
         let new_ptr = libc::realloc(ptr as *mut libc::c_void, new_size) as *mut u8;
 
-        ALLOC_LOCK.lock();
-        if ptr.is_null() {
-            // realloc is equivalent to malloc if called with a null pointer,
-            // thus we must insert the new block info in the ALLOC_INFO list.
-            ALLOC_INFO
-                .as_mut()
-                .unwrap()
-                .insert(ptr as usize, new_size, false);
-        } else {
-            // In all other cases, the memory block is resized and a new ptr
-            // is returned. This may differ from the original ptr if the new
-            // block is too large to fit in the same space as the old block
-            ALLOC_INFO
-                .as_mut()
-                .unwrap()
-                .update(ptr as usize, new_ptr as usize, new_size);
-        }
-        ALLOC_LOCK.unlock();
+        assert!(!ptr.is_null());
+        assert!(new_size > 0);
+
+        AllocMetadata::update(ptr as usize, new_ptr as usize, new_size);
         new_ptr
     }
 }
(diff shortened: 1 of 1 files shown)


src/alloc.rs
```diff
@@ -37,48 +37,22 @@ pub struct AllocWithInfo;
...
-            // In all other cases, the memory block is resized and a new ptr
-            // is returned. This may differ from the original ptr if the new
-            // block is too large to fit in the same space as the old block
-            ALLOC_INFO
-                .as_mut()
-                .unwrap()
-                .update(ptr as usize, new_ptr as usize, new_size);
-        }
-        ALLOC_LOCK.unlock();
+        debug_assert!(!ptr.is_null());
```
ltratt: I must admit, I'd be inclined to make this an `assert`, because it's such a strong condition that we need satisfied.

src/alloc.rs
```diff
@@ -37,48 +37,22 @@ pub struct AllocWithInfo;
...
+        debug_assert!(new_size > 0);
```
ltratt: Can we not handle zero-sized blocks? Some `malloc`s can, and I think it's worth us doing so if it's not too annoying. Just because the user *asked* for zero bytes doesn't mean we can't allocate some actual memory behind the scenes (e.g. 1 word) to give them a distinct pointer.

src/alloc.rs
```diff
@@ -396,16 +365,18 @@ impl AllocMetadata {
         ptr_info
     }
 
+    /// Updates the metadata associated with an allocation. If the block was
```
ltratt: "If the block was moved" sounds a bit vague. I think this comment just means "Update the allocator's information so that whe block once pointed to by `ptr` is recorded as having moved to `new_ptr` and is now `size` bytes big." or similar?

ltratt:
There's a minor type in "Move all pointer metadata queries into single API" [a missing "a"]. But I'm not sure that's a great title anyway because I'm not sure what "single API" means. Maybe we want to title this something like "Move all functions associated with pointer metadata into one place"?

src/alloc.rs
jacob-hughes: glibc `realloc` allows us to pass `size` 0: it simply returns a `nullptr` or equivalent which is safe to pass to `free`. However, the spec for `realloc` in the Rust `GlobalAlloc` trait states that this is disallowed, and will result in undefined behaviour. Presumably because they want compatibility with multiple allocators. 

I'm not sure that this is necessarily a problem, users could use `Gc` on a ZST and it would still allocate a word for the `GcHeader`.

ltratt: > However, the spec for realloc in the Rust GlobalAlloc trait states that this is disallowed, and will result in undefined behaviour.

Aha, OK, although I don't like that decision personally, I think it's reasonable enough for us to follow it. So ignore my comment (but maybe make it an `assert`?).

src/alloc.rs
jacob-hughes: Okay

src/alloc.rs
jacob-hughes: Fixed: 7b1ea0

src/alloc.rs
jacob-hughes: Agreed, updated: e0c96b

jacob-hughes:
Yeah that sounds better. I'll update that message when I squash.

src/alloc.rs
```diff
@@ -364,9 +364,9 @@ impl AllocMetadata {
         ptr_info
     }
 
-    /// Updates the metadata associated with an allocation. If the block was
-    /// moved to a different location in memory, a new pointer can be provided
-    /// to overwrite the old.
+    /// Updates the metadata associated with an allocation so that the block the
+    /// block once pointer to by `ptr` is recorded as having moved to `new_ptr`
```
ltratt: "the block the block" I think?

ltratt:
I wondered if you've also had a look at the other `debug_assert`s here? Basically my reasoning is that `debug_assert` should enforce conditions *within* the library. `assert` (or, perhaps preferably, `panic`) should enforce conditions on *users* of the library.

src/alloc.rs
jacob-hughes: Oops!

jacob-hughes:
I hadn't thought about it that way, but that certainly makes more sense. Unfortunately `panic` is not possible inside alloc calls. I've updated the others to match.

ltratt:
What does an `assert` do when it fails? I assumed it `panic`d but presumably I'm wrong!

jacob-hughes:
Ah apologies I had misremembered. Panicking is okay, but with a restriction: we can't panic with an error message string in-case it decides to allocate. For that we must use the `write` syscall.

ltratt:
OK. I'm fine with just using `assert` for errors caused by external users for the time being then.

jacob-hughes:
Am I okay to squash? I've pushed fix commits

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #13, opened 2019-11-07
Implement basic sweep phase

jacob-hughes:
This phase is ran after the mark phase, and calls `dealloc` on gc
objects which were not marked black. Note that as of yet there is no
support for finalizer or drop semantics: non-trivial `T`s stored in
`Gc<T>` which perform their own heap allocations will leak memory.

We now also run `lang_tester` in two phases: 1) with TSAN; and 2)
under Valgrind. Some tests now explicitly require valgrind to run. For
example, if we want to check that some memory was definitely freed, we
can try to read it later and assert that valgrind caught a
use-after-free. In cases like this, we add a filter to lang_tester to
exclude tests from phase 1 of testing if the file name contains the
substring "valgrind".


src/lib.rs
```diff
@@ -77,6 +78,9 @@ impl<T> Gc<T> {
         }
     }
 
+    pub(crate) fn base_ptr_offset(&self) -> usize {
+        self.header().0.get() & !1
```
ltratt: I'm a bit confused by the `&!1` -- maybe because I'm not really sure what this function is supposed to be doing.

src/gc.rs
```diff
@@ -154,6 +157,35 @@ impl Collector {
         }
     }
 
+    fn enter_sweep_phase(&mut self) {
+        *self.state.lock().unwrap() = CollectorState::Sweeping;
+
+        for PtrInfo { ptr, size, gc } in AllocMetadata.iter() {
```
ltratt: I don't know what this does because `PtrInfo`'s fields aren't documented.

gc_tests/run_tests.rs
```diff
@@ -12,19 +12,42 @@ use std::{env, path::PathBuf, process::Command};
...
+#[macro_use]
+extern crate lazy_static;
+
 static CRATE_NAME: &'static str = "gcmalloc";
 
-fn main() {
-    let sanitizers: Vec<String> = match env::var("SANITIZERS") {
-        Ok(ref s) => s.split(';').map(|s| s.to_string()).collect(),
-        Err(_) => Vec::new(),
+lazy_static! {
```
ltratt: Is there a good reason for these to be `lazy_static`? Couldn't they just be values returned by normal functions? `lazy_static` has a compile-time and run-time overhead, so if we don't need it, it would probably be good to avoid it.

ltratt:
@jacob-hughes OK, this PR is just (but just!) beyond my ability to review because there's not enough documentation on core aspects. For example, what's the mark bit and where is it stored? [OK, this one I eventually guessed.] Some key structs (e.g. `PtrInfo`) and many functions don't have any sort of hints on them. Even a brief sentence would help a lot here.

Basically, if we don't document some of this stuff soon (and, probably, reference it in commits), I will have completely lost a handle on this, which might be good in the short term, but probably isn't in the long term ;)

jacob-hughes:
Okay. I'll document some of the key data structures as part of this PR.

jacob-hughes:
I've added some documentation to the key moving parts in the GC. Let me know if you want anything else documenting or things aren't clear.

src/alloc.rs
```diff
@@ -309,15 +309,19 @@ enum Block {
...
+///
+/// This information is stored for every single allocation. It is vital for the
+/// collector to know things such as: determining whether an arbitrary word in a
+/// program is a ptr; how to get to the beginning of a block from some inner
+/// ptr; and whether a block is managed by the collector.
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub struct PtrInfo {
+    /// The pointer to the beginning of the allocation block.
     pub ptr: usize,
+    /// The size of the allocation block.
```
ltratt: In bytes, I assume?

src/lib.rs
```diff
@@ -40,15 +40,51 @@ static mut GC_ALLOCATOR: GCMalloc = GCMalloc;
...
     objptr: *mut T,
 }
 
+/// A garbage collected value is stored with a 1 machine word sized header. This
+/// header stores important metadata used by the GC during collection. It should
+/// never be accessible to users of the GC library. 
+///
+/// The metadata inside the header contains the mark-bit used by the collector
+/// to determine the value's reachability. It also contains the offset (in
+/// increments of usize) from the `objptr` to the base of the allocation block.
```
ltratt: What does "increments of usize" mean?

src/lib.rs
```diff
@@ -40,15 +40,51 @@ static mut GC_ALLOCATOR: GCMalloc = GCMalloc;
 
 static mut COLLECTOR: Option<Collector> = None;
 
+/// A garbage collected pointer. 'Gc' stands for 'Garbage collected'.
 #[derive(Debug)]
 pub struct Gc<T> {
     objptr: *mut T,
 }
 
+/// A garbage collected value is stored with a 1 machine word sized header. This
```
ltratt: I assume this word has a specific bit pattern? Can we document the layout?

ltratt:
This looks a lot better!

src/gc.rs
ltratt: I now suspect this might be faster if written as:

```rust
for PtrInfo { ptr, size, ... } in AllocMetadata.iter().filter(|x| x.gc) {
  ...
}
```

since the compiler is more likely to work out that it can optimise the iteration. I might be wrong though!

src/lib.rs
jacob-hughes: I should reword this as "the offset from the `objptr to `baseptr` in machine words."

ltratt: Because this is unusual, I think you also need to add "in multiples of machine words" and then give an example (something like "e.g. if this has the value 8, then the base pointer is, on a 64-bit machine, 64 bytes before the object pointer"). [I think it's unusual for a good reason BTW!]

Do you have a check somewhere that this number always fits in the number of bits available to it? Just in case someone chucks us something aligned to (say) 128GiB or equally weird...

jacob-hughes: > Do you have a check somewhere that this number always fits in the number of bits available to it? Just in case someone chucks us something aligned to (say) 128GiB or equally weird...

I don't. I make the assumption that it fits. This isn't very safe, but it might be worth just documenting it as such because this will be ripped out immediately in the next PR and replaced with a `vptr` when finalizers get put in.

ltratt: OK, in which case documenting it now is OK.

src/gc.rs
jacob-hughes: Yeah I agree that's better. Fixed: 39f2f41538b00f52d58f183b2fe1d0f5526d3faa

gc_tests/run_tests.rs
jacob-hughes: Good point. Fixed: 5aecf56216da33a541e92c20bbc5b0af69d322ec

src/alloc.rs
jacob-hughes: Yes. Fixed: ac8201e1afce6cb544f5cc6fab81324b076c4ba0

src/lib.rs
jacob-hughes: Okay. See what you think of c484ee07086a94b97ecbe648796e987923e78155

src/lib.rs
jacob-hughes: Sure: c484ee07086a94b97ecbe648796e987923e78155

jacob-hughes:
Ready for re-review.

ltratt:
Apart from the question about `format`, this looks good!

jacob-hughes:
Cool, can I squash?

ltratt:
We still need to fix the `0..64` thing.

jacob-hughes:
Okay, I've "fixed" this. (Even though 0..n always reads as n being exclusive to me!)

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
Pushed fixes for rustfmt

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
Please squash.

ltratt:
I'm an idiot.

jacob-hughes:
lol

PR #14, opened 2019-11-20
Represent `Gc<T>` in memory as thin trait objects.

jacob-hughes:
This is PR 1/2 for proper drop semantics for the GC. I've split this part out as it was quickly becoming complex and unreviewable! 

Part 1 fixes the inherent unsafety of storing the offset to the `baseptr` in a `Gc`'s header. This is done by storing each `Gc<T>` in memory as a thin trait object, with the `vptr` in the header. See each individual commit for more details about how this works.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -251,7 +251,7 @@ impl Collector {
         }
     }
 
-    fn mark(&self, obj: Gc<i8>, colour: Colour) {
+    fn mark(&self, mut obj: Gc<i8>, colour: Colour) {
         match colour {
             Colour::Black => obj.set_mark_bit(self.black),
             Colour::White => obj.set_mark_bit(!self.black),
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,9 +1,10 @@
 use crate::{
     alloc::{AllocMetadata, PtrInfo},
-    Gc, GC_ALLOCATOR,
+    Gc, Trace, GC_ALLOCATOR,
 };
 use std::{
     alloc::{Alloc, Layout},
+    mem::{align_of_val, size_of_val, transmute},
     ptr::NonNull,
     sync::Mutex,
 };
@@ -13,6 +14,11 @@ static WORD_SIZE: usize = std::mem::size_of::<usize>(); // in bytes
 type Address = usize;
 
 type Word = usize;
+
+/// Use this type when we do not care about the contents of a Gc. We choose `u8`
+/// because it maps similarly to how C / C++ use char when dealing with raw
+/// bytes.
+pub(crate) struct OpaqueU8(u8);
 
 type StackScanCallback = extern "sysv64" fn(&mut Collector, Address);
 #[link(name = "SpillRegisters", kind = "static")]
@@ -176,7 +182,7 @@ impl Collector {
                 // object's header. This means that unlike regular allocations,
                 // `ptr` will never point to the beginning of the allocation
                 // block.
-                let obj = unsafe { Gc::from_raw(ptr as *const i8) };
+                let obj = unsafe { Gc::from_raw(ptr as *const OpaqueU8) };
                 if self.colour(obj) == Colour::Black {
                     continue;
                 }
@@ -205,15 +211,9 @@ impl Collector {
         *self.state.lock().unwrap() = CollectorState::Sweeping;
 
         for PtrInfo { ptr, .. } in AllocMetadata.iter().filter(|x| x.gc) {
-            let obj = unsafe { Gc::from_raw(ptr as *const i8) };
+            let obj = unsafe { Gc::from_raw(ptr as *const OpaqueU8) };
             if self.colour(obj) == Colour::White {
-                unsafe {
-                    let baseptr = (ptr as *mut u8).sub(obj.base_ptr_offset());
-                    GC_ALLOCATOR.dealloc(
-                        NonNull::new_unchecked(baseptr as *mut u8),
-                        Layout::new::<usize>(),
-                    );
-                }
+                self.dealloc(obj);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -161,33 +161,48 @@ impl<T> Gc<T> {
...
-        } else {
-            self.0.set(self.0.get() & !1);
+        GcHeader {
+            base_ptr_offset: (uoff as u64).into(),
+            mark_bit: white,
         }
     }
 }
 
+
```
ltratt: Random blank line?

src/gc.rs
```diff
@@ -23,6 +24,11 @@ type Address = usize;
 
 type Word = usize;
 
+/// Use this type when we do not care about the contents of a Gc. We choose `u8`
+/// because it maps similarly to how C / C++ use char when dealing with raw
+/// bytes.
+pub(crate) type Placeholder = u8;
```
ltratt: This might be slightly dangerous as `type` in Rust doesn't create a new type, but just an alias. That means that a user who has a `Gc<u8>` struct will find that they are exactly equivalent to `Gc<Placeholder>`. One way around this might be to make this a newtype of sorts, maybe just:

```rust
struct Placeholder {
    byte: u8
}
```

Provided that struct can't escape the crate, there's then no chance of a clash.

ltratt: Alternatively, maybe just embrace the `u8`ness and delete the `type` alias?

ltratt:
I think there might be a possible disadvantage to this scheme: I don't think we can ever optimise away `drop` calls in trait objects because we always have to make the virtual call (even though, I guess, it's often just an empty function).

I wonder if the problem is that we're assuming the type of a `Gc` always tells us its size? What happens if I want to allocate a `Gc` DST of a struct with some arbitrary sized last element? I don't know how those are represented -- maybe you can only get at them with a vtable trick for their dynamic size / alignment?

Sorry I'm being a bit vague -- I'm thinking out loud somewhat!

jacob-hughes:
>  I don't think we can ever optimise away drop calls in trait objects because we always have to make the virtual call (even though, I guess, it's often just an empty function).

I'm still not sure about this. `drop_in_place` takes a `*mut T` and calls `T`'s `drop` method at some point afterwards. If `T` is a trait object, (i.e. `&mut dyn Trace`), it fetches the size / alignment out of the vtable, and calls the function pointer at `vtable[0]`: `T`'s `drop` method. If, however, `T` _is not_ a trait object, then I believe `drop_in_place` will be monomorphised to the concrete `drop` method for `T`. The problem with this is that during the sweep phase, we don't know the type of `T` statically, and refer to it with a placeholder: `Gc<Placeholder>` so `drop` calls have to be dynamically dispatched.

There may indeed be a way that we can optimise this: objects could be added to the drop queue iff they have a `drop` function pointer in their vtable. The documentation of `drop_in_place` hints to me that rustc knows this internally somehow with the line: 

> Executes the destructor (if any) of the pointed-to value.

While a trait object's first vtable entry is _always_ the `drop` fn pointer, there might be a way to check if it's valid or not (e.g. it's null if `T` has no destructor). I haven't checked this yet. 

I think this was the main benefit for using a separate `Finalizer` trait for finalizer-like semantics on `Gc<T>`s. By using specialisation, we could know at sweep-time whether or not an object could be deallocated in-place, or had to be added to a finalization queue because it contained `drop` code which needed running. 

> What happens if I want to allocate a Gc DST of a struct with some arbitrary sized last element? I don't know how those are represented -- maybe you can only get at them with a vtable trick for their dynamic size / alignment?

Yes, if there is no way to know statically the size of a type, then the only way to get this information is to fish it out of the vtable at runtime. 

tl;dr: I think it's too early to say whether or not the current `drop` approach _can_ be optimised.

src/gc.rs
jacob-hughes: I will play with the newtype idea of `Placeholder` first, as I find it nicely self-documents the distinction between `Gc<Everything>` and some actual `Gc<u8>`.

ltratt:
> While a trait object's first vtable entry is always the drop fn pointer, there might be a way to check if it's valid or not (e.g. it's null if T has no destructor). I haven't checked this yet.

I was wondering this too. I'd forgotten the "if any" clause in the docs. It would be great if we can check for NULL (or similar) and not call the function at all. Not that it hugely matters for now.

I'm inclined to say that we should stick with the approach you suggest in this PR and we can reconsider in the future if necessary.

src/gc.rs
ltratt: Actually, I was wondering if it should be a zero-sized type, because then you wouldn't give the illusion of sized-ness to the outside world. In other words, just `struct Placeholder;` would work I think?

src/lib.rs
jacob-hughes: Fixed: c9ad095c78364ecdff93a58c0798b1d9ace7bb18

src/gc.rs
jacob-hughes: A lot of the pointer arithmetic code relies on the `Placeholder` type being 1 byte sized. For simplicity, I've decided to keep it the same size. Fixed: 77c9918badbba02c7dfc27a709e27e874f4041b0

jacob-hughes:
> I'm inclined to say that we should stick with the approach you suggest in this PR and we can reconsider in the future if necessary.

Cool, that works for me. Ready for re-review.

src/gc.rs
ltratt: > A lot of the pointer arithmetic code relies on the Placeholder type being 1 byte sized.

I hadn't realised that. Then we should probably find a better name for `Placeholder` that makes that clearer. `OpaqueU8` or similar perhaps?

jacob-hughes: Yeah I can't really think of a better name than that. Try now

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #16, opened 2019-11-22
Introduce vector-like type based on `RawVec` and use to store block metadata 

jacob-hughes:
There were several problems with the old `AllocList` used to store block metadata: it couldn't grow in size; the element type was hard-coded; and it couldn't be statically initialized. This PR introduces a new vector-like type, `MaybeVec<T>`, based on the `RawVec` low-level utility. It's allocated using the standard allocator, which means it's ideal for use inside the collector as using it doesn't add or remove things from the block metadata list (as it would with the global allocator). This will be very useful in future when storing objects in a finalizer queue, as we won't accidentally resurrect objects.

Switching the `BLOCK_METADATA` list over to use `MaybeVec<T>` as its backing store means we can delete a large chunk of old `AllocList` related code.


src/alloc.rs
```diff
@@ -65,6 +67,119 @@ impl AllocLock {
...
+    /// it to the left, removal does not decrease the length of the vector.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index` is out of bounds.
+    pub fn remove(&mut self, index: usize) -> Option<T> {
+        assert!(index < self.len());
+        unsafe {
+            // infallible
+            let ret;
```
ltratt: I don't think the inner block does anything here -- the `unsafe` block itself defines a scope of equivalent length?

ltratt:
AFAICS `MaybeVec` never reuses elements after they've been deleted? I'm also slightly curious -- why the `Maybe` in `MaybeVec`?!

jacob-hughes:
> AFAICS MaybeVec never reuses elements after they've been deleted?

That's correct. This is not why I've done it this way. If I'm being honest, the main reason is that it's the easiest way do removal in a `Vec`-like interface. The problem with a `remove` method which shifts remaining elements to the left is that it can't be called in-place during iteration, which is a killer for performance when used inside the GC. In `Vec`, you get around this with methods like `drain` and `retain`. These are hard to get right and

jacob-hughes:
> why the Maybe in MaybeVec?!

Because naming is hard :( It's supposed to represent that there _might_ be a value at some `idx` (where `idx < vec.len()`). Unlike `Vec`, where this is guaranteed, `MaybeVec<T>` actually provides an `Option<T>` during access. 

Clearly, since you had to ask, the name doesn't really do the job ;) Perhaps `OptionVec` or `ShrinklessVec` or something? It's hard to come up with something concise which is still meaningful.

src/alloc.rs
jacob-hughes: This was inspired by `Vec`'s `remove` method which I chopped and changed in order to *not* shift. I'm guessing this is most likely a hangover from pre-NLL Rust. Simplified here: 3d7f208a62b8a86334878604d85d676bc4277712

ltratt:
> The problem with a remove method which shifts remaining elements to the left is that it can't be called in-place during iteration, which is a killer for performance when used inside the GC.

That suggests that using iteration is the wrong thing to do here. It feels like you've sort-of got a version of a linked list, except that your one can't ever reclaim memory? That might be very unfair of me, so feel free to push back!

jacob-hughes:
> That suggests that using iteration is the wrong thing to do here

Using iteration _is_ wrong if you want to call `remove(idx)` in a for loop. Using iteration to remove is normal in `Vec`s though using the `drain` and `retain` iter methods. Since I've not implemented the latter, I'm kind of stuck if I want to remove lots of elements in a single pass (which is what I do in the finalizer queue).

> It feels like you've sort-of got a version of a linked list,

It's a bit better than a linked-list. It's _mostly_ contiguous and each node doesn't require a separate heap allocation. 

> except that your one can't ever reclaim memory?

That's true for now. In a latter PR a `compact` function will be added to remove the holes in the list, but it wasn't necessary right now.

In essence, the behaviour of this vector is very similar to the previous `AllocList` implementation. The key difference is that unlike `AllocList`, which couldn't grow (leaving programs to crash if you overran its crude limit), `MaybeVec` grows infinitely. Using `RawVec` as the backing store gives us a bit more confidence too as Rust's existing collections rely on it. I think doing it this way too makes those trade-offs more obvious, and makes it easier to tweak them later if they're not quite right.

jacob-hughes:
Another key point is that having a collection like `MaybeVec` which is generic over some type `T` is really useful for building internal GC collections which need to make sure they don't accidentally keep objects alive. Since `MaybeVec`s allocator is completely separate to the global allocator, we can be sure of this. 

The original `AllocList` implementation couldn't easily be ported to some other use like, for example, the finalizer queue.

ltratt:
Wouldn't it be faster to make GC objects a linked list? Deleting an element would then be O(1) and there wouldn't be compaction issues.

ltratt:
On reflection, although I think this might not the perfect representation, it probably is better than what we had. Please squash.

jacob-hughes:
Before I squash, do you have any thoughts on a better name than `MaybeVec`? Perhaps `OptionVec` or `ShrinklessVec`?

ltratt:
Something like `HoleyVec` or `VecOfHoles` (or `VOH`) might be better?

jacob-hughes:
I've renamed  `MaybeVec` to `VOH`, let me know if you're happy and I'll squash.

ltratt:
The last commit made me laugh ;)

Please squash.

jacob-hughes:
Squashed!

PR #17, opened 2019-11-25
Add drop phase to GC

jacob-hughes:
Previously, the sweep phase would deallocate objects in-place, potentially causing leaks as their `drop` method was not called. In this PR, the collector queues dead objects for destruction while sweeping. It then enters the drop phase, where it iterates over the drop queue calling each object's destructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,47 @@
 # gcmalloc
 
 A Rust allocator with support for conservative garbage collection.
+
+## The Drop Trait
+
+`Gc` values implement the `Drop` trait. Unlike non-garbage-collected values,
+however, the `drop` method is not called when a `Gc` value goes out of scope.
+Instead, once the collector has determined that a value is garbage, it is added
+to a drop queue, where its `drop` method is ran at an indeterminate point in the
+future. This means that `Drop` now acts as both a regular destructor in the
+presence of scoped values, and a garbage collection finalizer when applied to
+`Gc` values.
+
+As with regular destructors, `drop` methods on `Gc` values are run from the
+"outside-in". However, if there are cycles, there is no way to know which `drop`
+method is safe to run first. Consider the example below:
+
+```rust
+
+struct Node<T> {
+    edge: Option<Gc<T>>
+}
+
+impl<T> Drop for Inner<T> {
+    fn drop(&mut self) {
+        println!("Points to: {}", self.edge.unwrap())
+    }
+}
+
+let x = Gc::new(Node { edge: None })
+let y = Gc::new(Node { edge: None })
+
+x.edge = Some(y);
+y.edge = Some(x);
+
+```
+
+Since an object is deallocated after its `drop` method is run, there is no safe
+order to run these destructors. If `x`'s `drop` method is called first, dropping
+`y` will dereference a dangling pointer. The same is true in the inverse order.
+
+To solve this, we impose an additional restriction on the `Drop` trait: you must
+not dereference a `Gc` inside the `drop` method. This works because unsafe
+`Drop` cycles can only be created using `Gc`s. By forbidding `drop` from
+derefing through `Gc`s, we can ensure `drop` only has access to resources which
+are guaranteed to be live.
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,5 +1,5 @@
 use crate::{
-    alloc::{AllocMetadata, PtrInfo, BLOCK_METADATA},
+    alloc::{AllocMetadata, PtrInfo, BLOCK_METADATA, VOH},
     GcBox,
 };
 
@@ -37,6 +37,7 @@ pub(crate) enum CollectorState {
     RootScanning,
     Marking,
     Sweeping,
+    Finalization,
 }
 
 /// Flags which affect the garbage collector's behaviour. They are useful for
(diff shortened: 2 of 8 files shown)


src/gc.rs
```diff
@@ -210,10 +210,12 @@ impl Collector {
     fn enter_sweep_phase(&mut self) {
         *self.state.lock().unwrap() = CollectorState::Sweeping;
 
-        for block in unsafe { BLOCK_METADATA.iter().filter(|x| x.map_or(false, |x| x.gc)) } {
-            let obj = unsafe { Gc::from_raw(block.unwrap().ptr as *const OpaqueU8) };
-            if self.colour(obj) == Colour::White {
-                self.dealloc(obj);
+        unsafe {
```
ltratt: Is everything about this `unsafe` including `iter`/`filter`? If not, can we make the `unsafe` part smaller?

src/lib.rs
```diff
@@ -91,72 +91,48 @@ pub struct Gc<T> {
...
-    /// the header.
-    fn header(&self) -> GcHeader {
-        let raw_bytes = unsafe {
-            let headerptr = (self.objptr as *const usize).sub(1) as *mut [u8; 8];
-            std::ptr::read(headerptr)
-        };
-        GcHeader::unpack(&raw_bytes).unwrap()
-    }
+/// A `GcBox` is a 0-cost wrapper which allows a single `Drop` implementation
+/// while also permitting multiple, copyable `Gc` references. The `drop` method
```
ltratt: Is it just me or is there not actually a `drop` method? I guess you mean something like "Note that `GcBox` intentionally does not define its own `drop` method, since dropping it should do nothing: only when the GC later determines that nothing points to the underlying object will that object's `drop` method be called."?

src/lib.rs
```diff
@@ -218,6 +198,10 @@ impl<T> DerefMut for Gc<T> {
     }
 }
 
+impl<T> Drop for GcBox<T> {
```
ltratt: Shouldn't this have been in the previous commit?

ltratt:
Modulo some minor comments, this looks good.

One question (that we talked about last week). If you have a cyclic data-structure then one object will be dropped first but I think that will allow the non-dropped thing to deref to the dropped thing and, presumably, segfault (or equivalent). That suggests to me that `Gc::deref` (or `GcBox::deref` or similar) might need to check the `dropped` flag and panic if it's `true`?

src/gc.rs
jacob-hughes: How about this? 813067bdda01ed6e7031ddbbfe7eeb118176dede

src/lib.rs
jacob-hughes: There is a `Drop` method, but as you noticed below I included it as part of the wrong commit ;). I'll fix this as part of squashing.

jacob-hughes:
> If you have a cyclic data-structure then one object will be dropped first but I think that will allow the non-dropped thing to deref to the dropped thing and, presumably, segfault (or equivalent). That suggests to me that Gc::deref (or GcBox::deref or similar) might need to check the dropped flag and panic if it's true?

Yes, you should never deref through a `Gc` in a `drop` method as it could cause a dangling pointer in the case of cyclic destructors. I'll make a PR which explains these rules in a README.

ltratt:
>Yes, you should never deref through a Gc in a drop method as it could cause a dangling pointer in the case of cyclic destructors. I'll make a PR which explains these rules in a README.

Can we not add a README with just that little bit of information to this PR (so that it's not forgotten)?

jacob-hughes:
I've added a basic explanation to the README. See what you think.

README.md
```diff
@@ -1,3 +1,40 @@
 # gcmalloc
 
 A Rust allocator with support for conservative garbage collection.
+
+## The Drop Trait
+
+Destructors for `Gc`s will be ran from the "outside-in", this guarantee is given
```
ltratt: I'm not sure that the clause "this guarantee is given to us..." is correct (surely we have to do some work to make it happen?) or useful. It might be easier to say "As with normal `drop`, `Gc` destructors are run from the "outside-in"."?

ltratt:
Interestingly, I *think* I have a use-case where I would like to access a `Gc` from a `Gc`'s `drop` in a cycle. I'm not sure yet though. If true, I suspect we'll be able to do something like provide a method `Gc::see_if_dropped(...) -> Option<Gc<T>>` which given a reference to a `Gc<T>` returns `Some` if the thing hasn't been dropped yet and `None` otherwise. I don't think we need to handle it yet, though, because I might be very wrong!

README.md
jacob-hughes: I've re-worded this and added an opening paragraph for context. I'm wary not to use the term _destructor_ interchangeably with _finalizer_ as GC terminology does make them distinct.

jacob-hughes: I'm also aware that I should be spelling finalizer is with an _s_! :gb: But Java broke me from an early age and anything other than the American spelling looks weird to me.

ltratt:
Please squash.

README.md
ltratt: It's OK, "finaliser" is also acceptable British English, so you don't have to worry about betraying the Queen ;)

jacob-hughes:
Squashed

PR #18, opened 2019-12-03
Track allocations using a linked-list

jacob-hughes:
This PR is going to be a bit difficult to review as it touches every part of the
collector. I've done my best to split the commits chronologically.

During collection, we need a way to traverse each heap allocated block
found from the rootset for further pointers. Previously this was done
using a global vector which stored metadata about each allocation. This
was brittle and lead to synchronisation problems.

This commit adds a header to each block with pointers to next and
previous blocks, allowing them to traversed during collection. A doubly
linked list is used so that deallocation is O(1) instead of O(n).

We also fold the header for GC objects up into the block header.
Although we take a memory hit in doing this with an extra 4 machine
words necessary per each allocation, the implementation is simplified
and easier to debug in the meantime.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -248,32 +248,20 @@ impl Iterator for BlocksIter {
     }
 }
 
-/// A VOH (Vector of Holes) is a contiguous growable array type optimised for
-/// fast insertion and O(1) removal.
+/// A GcVec is a contiguous growable array type whose contents are not tracked
+/// for garbage collection.
 ///
-/// Unlike the `Vec<T>` type available in the standard library, `VOH<T>`
-/// does not shift remaining elements to the left after removing an item.
-/// Instead, a `None` value is swapped in its place, effectively leaving a hole
-/// inside the vector.
-///
-/// Accessing elements in a `VOH<T>`, however, can be slower than a
-/// `Vec<T>`, as elements are no longer strictly contiguous; fragmentation will
-/// occur over time as removed elements increase the distance between those
-/// which remain.
-///
-/// The API to `VOH<T>` can be thought of as a heavily stripped down
-/// version of `Vec<T>`. Its use-case is highly specific, and intended only to
-/// be used internally as part of the Collector implementation. For this reason,
-/// its allocation is not tracked by the collector. Do not store any values
-/// inside which you intend the GC to keep alive.
-pub(crate) struct VOH<T> {
-    buf: RawVec<Option<T>, System>,
+/// The API to `GcVec` can be thought of as a heavily stripped down version of
+/// `Vec`. It is intended only to be used internally as part of the Collector
+/// implementation where objects do not need keeping alive by the GC.
+pub(crate) struct GcVec<T> {
+    buf: RawVec<T, System>,
     len: usize,
 }
 
-impl<T> VOH<T> {
+impl<T> GcVec<T> {
     #[inline]
-    pub const fn new() -> VOH<T> {
+    pub const fn new() -> GcVec<T> {
         Self {
             // use the system allocator instead of the global allocator
             buf: RawVec::new_in(System),
@@ -294,7 +282,7 @@ impl<T> VOH<T> {
         }
         unsafe {
             let end = self.as_mut_ptr().add(self.len);
-            ::std::ptr::write(end, Some(value));
+            ::std::ptr::write(end, value);
             self.len += 1;
         }
     }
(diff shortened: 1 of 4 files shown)


src/alloc.rs
```diff
@@ -25,29 +27,160 @@ pub(crate) static mut BLOCK_METADATA: VOH<PtrInfo> = VOH::new();
...
+pub(crate) struct BlockHeader {
+    prev: *mut BlockHeader,
+    next: *mut BlockHeader,
+    size: NonZeroUsize,
+    pub(crate) type_: BlockType,
+}
+
+#[derive(Eq, PartialEq)]
+pub(crate) enum BlockType {
+    Gc([u8; 7]),
```
ltratt: Is this intentionally 7 bytes not 8?

src/gc.rs
```diff
@@ -201,8 +196,11 @@ impl Collector {
             for addr in (ptr..ptr + size).step_by(WORD_SIZE) {
                 let word = unsafe { *(addr as *const Word) };
 
-                if let Some(ptrinfo) = AllocMetadata::find(word) {
-                    self.worklist.push(ptrinfo)
+                if let Some(block) = ALLOCATOR
+                    .iter()
+                    .find(|x| x.ptr == word || word > x.ptr && word < x.ptr + x.size)
```
ltratt: Should this be `word >= x.ptr`?

src/gc.rs
```diff
@@ -288,8 +286,11 @@ impl Collector {
 
         for stack_address in (rsp..stack_top).step_by(WORD_SIZE) {
             let stack_word = unsafe { *(stack_address as *const Word) };
-            if let Some(ptr_info) = AllocMetadata::find(stack_word) {
-                self.worklist.push(ptr_info)
+            if let Some(block) = ALLOCATOR
+                .iter()
+                .find(|x| x.ptr == stack_word || stack_word > x.ptr && stack_word < x.ptr + x.size)
```
ltratt: Ditto.

src/gc.rs
jacob-hughes: It can be, I don't think it matters really because if `word == x.ptr` the LHS of the or expression should short circuit.

jacob-hughes:
I've pushed a commit that re-orders the `BlockHeader` layout to contain Gc header information in a less brittle format. Sorry it's quite large and will be a bit of a pain to review.

src/gc.rs
ltratt: Ah, yeah.

src/alloc.rs
```diff
@@ -94,74 +82,113 @@ unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
...
-    pub(crate) type_: BlockType,
+    metadata: [u8; SIZE_BLOCKMETADATA],
+}
+
+const SIZE_BLOCKMETADATA: usize = 16;
+
+#[derive(PackedStruct, Default, Debug)]
+#[packed_struct(bit_numbering = "msb0")]
+pub struct BlockMetadata {
+    /// The size of the block (excluding its header)
```
ltratt: Let's consistently add "." at the end of these comments. Petty, I know :)

src/alloc.rs
```diff
@@ -94,74 +82,113 @@ unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
...
+    #[packed_field(size_bits = "62", endian = "msb")]
```
ltratt: There's an implicit assumption here that the world is a 64-bit integer. I personally like to add some sort of guard to every such assumption so that if/when we move to different word sizes every part of the program that needs to be reconsidered is explicitly flagged.

ltratt:
I have no substantial comments, and just a few nitpicks.

That said, having first seen this yesterday and been unsure about it, I'm not sure that having variables called `ptr` and then explicitly saying `::ptr::read` and so on is ideal -- I'm too used to `ptr` being the standard module! Maybe find another variable name (e.g. `p`) and then you can simplify things accordingly?

src/alloc.rs
jacob-hughes: I agree with this in principle. There already exists a global check on the word size inside the library. Are you suggesting to duplicate this check to places where we make implicit assumptions that a word is 64-bit?

ltratt: Yep.

jacob-hughes: Fixed: 1bd2ee26a71cc4e1aaa443eeaba44d4ee7c90cf4

src/alloc.rs
jacob-hughes: Fixed: 4c15d2aba7d080e04742f0736644ad5c05962f49

jacob-hughes:
Ready to squash if you're happy.

ltratt:
I like https://github.com/softdevteam/gcmalloc/commit/513afc5ebbce56f94b0c7dfda27ebf2a7f0ee136 but can't you then replace all `::std::ptr::`s with just `ptr::`?

jacob-hughes:
Ah I see what you mean now. Fixed: 27ded925f4883aa7d2d0c26dd8749289a9b96108

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
The last commit needs a bit of rephrasing in the commit title/message: it doesn't make VOH contiguous, so much as completely rethink and rename it :)

jacob-hughes:
Fair point. See what you think of this instead.

ltratt:
Works for me!

bors r+

PR #19, opened 2019-12-06
Add fully static initialization of collector

jacob-hughes:
This should make the GC easier to integrate with existing projects, as
we remove the requirement to manually run `init` in main. Note that we
switch to using the `parking_lot` synchronisation primitives as they
don't require heap-allocation and their constructors are const
functions.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,13 +1,12 @@
 use crate::{
     alloc::{GcVec, PtrInfo},
-    GcBox, ALLOCATOR, GC_ALLOCATOR,
+    GcBox, ALLOCATOR, COLLECTOR_STATE, GC_ALLOCATOR,
 };
 
 use std::{
     alloc::{Alloc, Layout},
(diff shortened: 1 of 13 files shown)


src/gc.rs
```diff
@@ -41,18 +40,32 @@ pub(crate) enum CollectorState {
     Finalization,
 }
 
+impl CollectorState {
+    fn update(&mut self, new_state: CollectorState) {
+        let err = format!("Invalid GC state transition: {:?} -> {:?}", self, new_state);
```
ltratt: Since only one arm of the `match` references `err`, shouldn't we move it there?

jacob-hughes: Sure. Fixed: 3b68497defa403765875a62aa3bc1c0699bf051e

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #21, opened 2019-12-06
Add a basic allocation counter to trigger GC

jacob-hughes:
Until now, garbage collection could only be triggered manually (via
`gcmalloc::collect()`). This adds a simple counter which increments on
the allocation of new `Gc` values until a threshold has been reached.
This arbitrarily defaults to 100 `Gc` values.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -7,6 +7,8 @@ use std::{
     sync::atomic::{AtomicBool, Ordering},
 };
 
+use crate::COLLECTOR;
+
 use packed_struct::prelude::*;
 
 /// A spinlock for keeping allocation thread-safe.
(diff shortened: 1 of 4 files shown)


src/alloc.rs
```diff
@@ -40,6 +42,8 @@ unsafe impl GlobalAlloc for GlobalAllocator {
 
 unsafe impl Alloc for GcAllocator {
     unsafe fn alloc(&mut self, layout: Layout) -> Result<NonNull<u8>, AllocErr> {
+        COLLECTOR.lock().poll();
```
ltratt: I sort-of assumed we only want to `lock` once? Admittedly, I can't remember if unlocking is automatic or not...

jacob-hughes: Yeah I agree it's not ideal to lock the collector to poll it, and then immediately lock it again afterwards to bump `allocations`. Let me rewrite this with a temporary so we acquire the lock only once.

jacob-hughes: Fixed: ce0fa64d22f53af694df517f2b46b02c94a742c0

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #22, opened 2019-12-13
Scan for roots in static variables

jacob-hughes:
Since statics can be mutable, they may contain roots to garbage
collected values. We identify the range where these roots can live in
the .data segment and scan it for pointers in a similar manner to the
stack. This is platform specific and we only provide a mechanism for
doing this in linux at the moment.

### DIFF
--- a/gc_tests/valgrind.supp
+++ b/gc_tests/valgrind.supp
@@ -2,12 +2,6 @@
    <uninitialized_stack>
    Memcheck:Cond
    ...
-   fun:scan_stack
+   fun:_ZN8gcmalloc2gc9Collector13check_pointer*
 }
 
-{
-   <insert_a_suppression_name_here>
-   Memcheck:Cond
-   fun:_ZN8gcmalloc2gc9Collector16enter_mark_phase*
-   ...
-}
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -23,6 +23,8 @@ static ALLOC_LOCK: AllocLock = AllocLock::new();
 
 static mut LAST_ALLOCED: *mut BlockHeader = ::std::ptr::null_mut();
 
+pub static mut HEAP_TOP: usize = 0;
+
 pub struct GlobalAllocator;
 pub struct GcAllocator;
 
@@ -78,6 +80,11 @@ unsafe fn alloc(layout: Layout, is_gc: bool) -> *mut u8 {
     ::std::ptr::write(headerptr, header);
     LAST_ALLOCED = headerptr;
 
+    let top = p.add(layout.size());
+    if top as usize > HEAP_TOP {
+        HEAP_TOP = top as usize;
+    }
+
     ALLOC_LOCK.unlock();
     p
 }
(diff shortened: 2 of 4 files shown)


gc_tests/valgrind.supp
```diff
@@ -2,12 +2,6 @@
    <uninitialized_stack>
    Memcheck:Cond
    ...
-   fun:scan_stack
+   fun:_ZN8gcmalloc2gc9Collector13check_pointer*
```
ltratt: I don't know the mangling rules, but I was sort of expecting something like `*Collector*_pointer` or similar? Obviously I can see problems with the wildcards, so maybe it's a bad idea!

src/alloc.rs
```diff
@@ -23,6 +23,8 @@ static ALLOC_LOCK: AllocLock = AllocLock::new();
 
 static mut LAST_ALLOCED: *mut BlockHeader = ::std::ptr::null_mut();
 
+pub static mut HEAP_MAX: usize = 0;
```
ltratt: `HEAP_MAX` sounds like it's related to the size of the heap: `HEAP_HIGH` might be clearer?

And then shouldn't we be tracking both `HEAP_LOW` and `HEAP_HIGH` not just the latter?

jacob-hughes: Maybe `HEAP_TOP`? As it's similar to our descriptions of the stack. 

We don't need to track the lower bound of the heap because it always starts at the end of `.bss` and the linker already gives us a symbol we can use to get that address.

ltratt: Works for me.

> We don't need to track the lower bound of the heap because it always starts at the end of .bss

Aha, can we document that somewhere?

jacob-hughes: Sure. Fixed: ab74aea33982c5397306b6a45e83d4991ab5699a

src/gc.rs
```diff
@@ -335,7 +335,9 @@ impl Collector {
     }
 
     fn check_pointer(&mut self, word: Word) {
-        if word > self.data_segment_end && word < unsafe { crate::alloc::HEAP_MAX } {
+        // Since the heap starts at the end of the data segment, we can use this
+        // as the lower heap bound.
+        if word > self.data_segment_end && word < unsafe { crate::alloc::HEAP_TOP } {
```
ltratt: Should this be `if word >= self.data_segment_end`?

We also might need to do `word <= unsafe { crate::alloc::HEAP_MAX }` because we might need to support the C idiom whereby you can safely move a pointer to the first byte after a block (at which point we should probably consider such a pointer to be within that block). Thoughts?

jacob-hughes: > Should this be if word >= self.data_segment_end?

Ouch, yes you're right.

> We also might need to do word <= unsafe { crate::alloc::HEAP_MAX }

You have read my mind. I actually have another PR coming up to deal with exactly this issue. While what you said is true, there is more to it as well. We need the marking phase to deal with this explicitly also. If we have a pointer directly to the first address in a block, we need to mark both that block and the previous block as there's no way of telling which block the mutator is trying reach.

ltratt: OK, let's fix the first thing in this PR and the second thing in the next PR.

jacob-hughes: Fixed: f3ddd89

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #23, opened 2019-12-16
Keep Gc value alive if pointed to by one-element-after ptr

jacob-hughes:
Rust (and C) allow pointers 1 byte after the end of an array providing
it is never dereferenced. The GC must take this into account and keep
values with one-element-after pointers alive.

### DIFF
--- a/gc_tests/tests/c_pointer_after.rs
+++ b/gc_tests/tests/c_pointer_after.rs
@@ -1,0 +1,34 @@
+// Run-time:
+//  status: success
+
+extern crate gcmalloc;
+
+use gcmalloc::{collect, gc::DebugFlags, Debug, Gc};
+
+#[inline(never)]
+fn setup_obj() -> Box<*mut u8> {
+    let a = Gc::new(123_u8);
+
+    // Return a pointer to 1 byte past the end of the value. This is valid in
+    // the C (and Rust) standard, so it should keep the object alive.
+    //
+    // We box it to prevent the original pointer rooting the object through a
+    // stale register across the call
+    Box::new(unsafe { a.as_ptr().add(1) as *mut u8 })
+}
+
+#[inline(never)]
+fn check_marked(ptr: Box<*mut u8>) {
+    // Reconstruct the base pointer to see if it was marked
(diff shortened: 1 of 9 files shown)


gc_tests/tests/c_pointer_after.rs
```diff
@@ -0,0 +1,36 @@
...
+
+fn main() {
+    gcmalloc::debug_flags(DebugFlags::new().sweep_phase(false));
+
+    let a = setup_obj();
+    collect();
+
+    check_marked(a);
+}
+
```
ltratt: Random blank lines.

src/gc.rs
```diff
@@ -337,10 +337,13 @@ impl Collector {
...
         // as the lower heap bound.
-        if word >= self.data_segment_end && word < unsafe { crate::alloc::HEAP_TOP } {
+        if word >= self.data_segment_end && word <= unsafe { crate::alloc::HEAP_TOP } {
             if let Some(block) = ALLOCATOR
                 .iter()
-                .find(|x| x.ptr == word || word > x.ptr && word < x.ptr + x.size)
+                // It's legal to hold a pointer 1 byte past the end of a block,
+                // but we can still use find because this will never over-run
+                // the size of the next block's header.
+                .find(|x| x.ptr == word || word > x.ptr && word <= x.ptr + x.size + 1 )
```
ltratt: I think the `+ 1` at the end is wrong: chaging `<` to `<=` is sufficient. The `+1` then allows a pointer to *2 bytes* after the end, which I don't think we should support.

jacob-hughes: Bah! You're right. I was only testing you...

gc_tests/tests/c_pointer_after.rs
jacob-hughes: Fixed: 26fde82502b35e9e670fc3d270937352e402f91a

src/gc.rs
jacob-hughes: Fixed: a6dd3254eab3480dab7506851d0d199e70077640

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #24, opened 2019-12-17
Add basic support for `Gc<T: ?Sized>`

jacob-hughes:
In order to have the same flexibility as other smart pointer types, it
should be possible for users to store DSTs inside a `Gc` container. To do
this, we use the magic `CoerceUnsized` trait to convert a concrete type
to an unsized one when requested by the user at a "coercion point".

For example, lets imagine we have a trait `MyTrait` and a struct `S`,
where `S: MyTrait + Sized`. We might want to store a garbage collected
reference to a bare trait object as below:

    let gc: Gc<dyn MyTrait> = Gc::new(s) // where `s: S`.

This now compiles successfully because of our `CoerceUnsized`
implementation on `Gc`. It works because the compiler can see a valid
coercion from a concrete type to a DST: more specifically, the concrete
type `S` can be converted to the trait object `dyn MyTrait`, because `S`
is bound by the trait `MyTrait`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,8 @@
 #![feature(alloc_layout_extra)]
 #![feature(raw_vec_internals)]
 #![feature(const_fn)]
+#![feature(coerce_unsized)]
+#![feature(unsize)]
 #[cfg(not(all(target_pointer_width = "64", target_arch = "x86_64")))]
 compile_error!("Requires x86_64 with 64 bit pointer width.");
 
@@ -23,8 +25,9 @@ use crate::{
 };
 use std::{
     alloc::{Alloc, Layout},
+    marker::Unsize,
     mem::{forget, transmute, ManuallyDrop},
-    ops::{Deref, DerefMut},
+    ops::{CoerceUnsized, Deref, DerefMut},
 };
 
 use parking_lot::Mutex;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -248,3 +258,33 @@ impl Debug {
...
+        struct HelloWorldStruct(usize);
+
+        impl HelloWorld for HelloWorldStruct {
+            fn hello(&self) -> usize {
+                self.0
+            }
+        }
+
+        let s = HelloWorldStruct(123);
+        let gcto: Gc<dyn HelloWorld> = Gc::new(s);
```
ltratt: I think you probably need two `impl`s for this struct, otherwise Rust might end up turning this into static dispatch. Alternatively, you could assert that `size_of::<Gc<dyn HelloWorld>>()` is `2 * usize`?

jacob-hughes: > Alternatively, you could assert that size_of::<Gc<dyn HelloWorld>>() is 2 * usize?

Good idea. I'll add this.

jacob-hughes: Fixed: 75d151976e9ab0f6b8883c38c719450b3978877c

ltratt:
bors try

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #27, opened 2019-12-18
Fix bug where realloc did not update block size

jacob-hughes:
This caused the contents of vectors not to be fully traced if their
capacity grew over time, leading to the freeing of live objects. Also
simplify the test for this and ensure it is no longer ignored.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -111,13 +111,17 @@ unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
 unsafe fn realloc(ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
     ALLOC_LOCK.lock();
     let old_hp = (ptr as *mut BlockHeader).sub(1);
-    let header = ::std::ptr::read(old_hp);
+    let mut header = ::std::ptr::read(old_hp);
 
     let (l2, uoff) = Layout::new::<BlockHeader>().extend(layout).unwrap();
     let p = System.realloc(ptr.sub(uoff), l2, new_size + uoff).add(uoff);
 
     let new_hp = (p as *mut BlockHeader).sub(1);
-    header.metadata().size = (new_size as u64).into();
+    let mut md = header.metadata();
+
+    assert_eq_size!(usize, u64);
+    md.size = (new_size as u64).into();
+    header.set_metadata(md);
 
     BlockHeader::patch_next(header.prev, new_hp);
     BlockHeader::patch_prev(header.next, new_hp);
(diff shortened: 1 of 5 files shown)


src/allocator.rs
```diff
@@ -111,13 +111,15 @@ unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
...
-    let header = ::std::ptr::read(old_hp);
+    let mut header = ::std::ptr::read(old_hp);
 
     let (l2, uoff) = Layout::new::<BlockHeader>().extend(layout).unwrap();
     let p = System.realloc(ptr.sub(uoff), l2, new_size + uoff).add(uoff);
 
     let new_hp = (p as *mut BlockHeader).sub(1);
-    header.metadata().size = (new_size as u64).into();
+    let mut md = header.metadata();
+    md.size = (new_size as u64).into();
```
ltratt: This cast slightly worries me: there's no reason that `usize` can safely be converted to a `u64`. And then I'm not really sure why we need `into` as well. What's the type of `size`?

jacob-hughes: In this case the cast is safe. `BlockMetadata.size` is a bit-packed `u64` represented with 62 bits (so a `u62` really). If `usize` were 32 / 16 bit, its size would be safely upcast to use 64 bits regardless of architecture. The `into()` converts the `u64` into a `packed_bits::Bits62`. 

See https://github.com/softdevteam/gcmalloc/blob/2f55d998b3179ffabc90fc66bf79376c3926da30/src/allocator.rs#L155

The reason we steal bits from `size` is because it's very unlikely that we'd have a layout where the size ends up being anywhere near `usize::MAX_INT`.

ltratt: This cast is only safe on machines with 64 bit pointers. So we need to have some sort of `const_assert` or whatever here, because this cast isn't safe in general.

jacob-hughes: Why? If we cast from a smaller integer ( say a machine where `usize` is 32bit) to a `u64` then it will just zero-extend. So the value of size won't change. The `BlockMetadata` will just end up storing some extra unused zeroes.

This would only be unsafe on architectures where `usize` > `u64`, as it would truncate on cast. Is that what you're referring to? I assumed this was so rare that we just won't support it (and I'm sure I've made several other assumptions elsewhere where this would break!)

jacob-hughes: I've added a fix that asserts `usize` is `u64` here for clarity: f1eacae9e3dc945d6843671585b076211b8bc96c

ltratt:
Please squash.

jacob-hughes:
Squashed.

PR #28, opened 2020-01-09
Add `Any` type support to Gc<T>

jacob-hughes:
The `Any` type provides a very restricted form of dynamic typing in
Rust. This means that a `Gc<T>` (where `T: Any`) can now be downcast to
its concrete type. `Any` also allows for testing whether values are of a
specified concrete type.

The `Any` trait does not, however, allow the user to see which traits
the `Gc<T>` implements.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -1,5 +1,6 @@
 use std::{
     alloc::{Alloc, Layout},
+    any::Any,
     marker::Unsize,
     mem::{forget, transmute, ManuallyDrop},
     ops::{CoerceUnsized, Deref, DerefMut},
@@ -45,6 +46,20 @@ impl<T> Gc<T> {
     pub fn new(v: T) -> Self {
         Gc {
             objptr: unsafe { NonNull::new_unchecked(GcBox::new(v)) },
+        }
+    }
+}
+
+impl Gc<dyn Any> {
+    pub fn downcast<T: Any>(&self) -> Result<Gc<T>, Gc<dyn Any>> {
+        if (*self).is::<T>() {
+            let ptr = self.objptr.cast::<GcBox<T>>();
+            forget(self);
+            Ok(Gc { objptr: ptr })
+        } else {
+            Err(Gc {
+                objptr: self.objptr,
+            })
         }
     }
 }


src/gc.rs
```diff
@@ -49,6 +51,18 @@ impl<T> Gc<T> {
     }
 }
 
+impl Gc<dyn Any> {
+    pub fn downcast<T: Any>(self) -> Result<Gc<T>, Gc<dyn Any>> {
```
ltratt: I'm not sure I see the utility in having `downcast` take `self` rather than `&self`. `downcast(self)` makes sense on a `Box`, for example, because you're also consuming the underlying memory, but on a `Gc` you're not. Thoughts?

ltratt: [For example, `Val::downcast` in yksom takes `&self` https://github.com/softdevteam/yksom/blob/master/src/lib/vm/val.rs#L130]

jacob-hughes: I did think this, but I actually think it makes sense to consume the `Gc<T>` wrapper as we return a new one with the more concrete type. This feels a bit more in-keeping with Rust's move semantics. While it's true that we implement `Copy` anyway. It feels a bit more explicit and means we won't have handles of different types lingering in scope unless we really ask for it. 

This is also the `Rc` approach btw, so it also means we would have similar APIs.

ltratt: I know this is how `Rc` does it but I haven't yet really worked out *why* `Rc` does it this way (I guess that `Rc` followed `Box` and I can see why `Box` does it this way). I am unable to see any benefits from `self` (vs. `&self`) and some disadvantages (sometimes `downcast` forces you to `clone` a `Gc` in order that you still have a reference to the non-downcasted version). Put another way, everything you can do with `self` you can do with `&self` but not vice versa AFAICS?

jacob-hughes: Yes that's right. I don't have very strong opinions on this, although I would personally lean further towards consistency. If you think it would make life easier as a consumer of `gcmalloc` for `Gc` to use `&self` then I don't mind changing it.

ltratt: Consistency is a laudable goal, but sometimes being inconsistent is more laudable :) Since we agree that `&self` is strictly more flexible -- and doesn't have any downsides that we know of -- I think that's the way to go.

jacob-hughes: Fixed: 70f15f73309e0c26cf3ea58cc597f6d3570e5143

jacob-hughes:
Ready for re-review. I've also removed an import which slipped through the net (e35f1b03057bb30fbc01d41323728b00477f95dc).

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
Oops. Rustfmt. Can I force-push an update?

ltratt:
You need to push a separate commit first I'm afraid.

jacob-hughes:
Ok. Fixed: 643f91e

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #30, opened 2020-01-13
Add wrapper type for a block pointer

jacob-hughes:
This provides a common interface for accessing and mutating block header
information. 

This is a precursor to some larger changes. Looking at this from a higher-level,
I've noticed a couple of bugs which spring up from assumptions made by the 
Collector about when an allocation block (and its metadata) is safe to dereference. 
Hopefully, this commit paves the way for a change where we can be more 
confident that certain accesses in GC land are safe, by taking advantage of Rust 
lifetimes.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -7,7 +7,7 @@ use std::{
     sync::atomic::{AtomicBool, Ordering},
 };
 
-use crate::COLLECTOR;
+use crate::{gc::Colour, COLLECTOR};
 
 use packed_struct::prelude::*;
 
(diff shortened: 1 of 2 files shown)


src/allocator.rs
```diff
@@ -218,6 +219,82 @@ impl BlockHeader {
...
+/// raw pointer and makes no guarnatees about the liveness of the block it
+/// represents. Use of the `Block` API without first establishing that the
+/// corresponding allocation block is live can lead to use-after-frees.
+///
+/// This API is safe to use on uninitialized blocks, as a Block's header is
+/// always constructed before a block pointer is returned to the application.
+///
+pub(crate) struct Block<'a> {
+    base: *mut u8,
+    phantom: PhantomData<&'a ()>,
```
ltratt: At the moment I don't think any of the code makes use of `'a` does it? i.e. if you removed this from the struct, I think this PR would still compile OK?

jacob-hughes: You're right. It will be necessary in future, but it doesn't make sense on its own. I've removed it from this PR (6368f279a728ae2a0dd67c874fd9a92372b7021d)

ltratt:
Please squash.

jacob-hughes:
Squashed

ltratt:
[Cancelled at @jacob-hughes request.]

jacob-hughes:
Pushed fmt fix

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #31, opened 2020-01-14
Use `Block` in collector with proper move semantics

jacob-hughes:
Previously, we used a struct called `PtrInfo` which held some
information about each block for various methods in the `Collector`.
`PtrInfo` was copyable, had public fields, and allowed us to retain
information about a block within the `Collector` methods even after they
had been freed. This was very unsafe as derefs to block pointers within
the collector were all based on hidden assumptions about how the GC
algorithm worked.

This commit attempts to minimise this unsafety by using move semantics
and information hiding to ensure the following when accessing blocks
within the Collector methods:

     

-  The actual block pointer is hidden from the collector, this
    prevents it from being derefed unless done through restricted
    APIs on the `Block` interface. When the collector needs to perform
    pointer comparisons with arbitrary words, these words are instead
    passed to the block with the result returned to the collector.

- The `Block` pointer interface is no longer copy/cloneable, which
    means that destructive operations such as `drop` can consume the `Block`
    to help prevent it being used later in the collector.

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -233,6 +233,7 @@ impl BlockHeader {
 /// This API is safe to use on uninitialized blocks, as a Block's header is
 /// always constructed before a block pointer is returned to the application.
 ///
+#[derive(Debug, PartialEq, Eq)]
 pub(crate) struct Block {
     base: *mut u8,
 }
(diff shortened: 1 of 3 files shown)


src/allocator.rs
```diff
@@ -288,8 +300,26 @@ impl Block {
         md.drop_vptr = (value as u64).into();
         self.header_mut().set_metadata(md);
     }
+
+    pub(crate) unsafe fn drop(self) {
+        assert!(
```
ltratt: I'm not totally sure I understand how/why this `assert` might be called.

jacob-hughes: Drop has a precondition that can only be called on allocation blocks which are managed by the GC. Since we've now moved `drop` to be a method call on `Block` instead of directly inside the `enter_drop_phase` method in the collector, it is theoretically possible to construct a `Block` from a base pointer to a non-garbage-collected allocation block. This assert acts as a sanity check against that.

ltratt: Can the end user ever do this or is it an internal sanity check? If the latter, should it be `debug_assert`?

jacob-hughes: It's an internal sanity check, the end user can never drop a `Gc` value explicitly. I'm happy to make it a `debug_assert`.

ltratt: OK, let's do that then.

jacob-hughes: Fixed: 966631df121a9fda0f79e7d189bc9ede62c29c51

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #32, opened 2020-01-15
Encapsulate unsafe heap access through explicit references on Block API.

jacob-hughes:
This PR attempts to make the `Collector` more ergonomic by using mostly
Rust references to heap blocks inside the Collector methods. Previously we 
were quite liberal in our use of raw pointers in the `Collector` methods. The only 
thing ensuring that these pointers were used correctly were following the 
invariants of a mark-sweep algorithm. These existed mostly in my
head (and of course, I had got this wrong in a few places!)

By using references with explicit lifetimes in this way, we can reduce exposing
raw pointers in the collector and try and put some of the burden of ensuring
correct usage on the borrow-checker. Part of the reason we need to be so
thorough with this is because it is completely UB to hold an invalid reference
in Rust (null, out-of-bounds, etc).

Please see each commit message for more details.

@vext01 I've added you as a reviewer as you may be interested because we
discussed this on Tuesday.

### DIFF
--- a/gc_tests/valgrind.supp
+++ b/gc_tests/valgrind.supp
@@ -2,6 +2,6 @@
    <uninitialized_stack>
    Memcheck:Cond
    ...
-   fun:_ZN8gcmalloc9collector9Collector13check_pointer*
+   fun:_ZN8gcmalloc9allocator5Block4find*
 }
 
--- a/src/SpillRegisters_X64.S
+++ b/src/SpillRegisters_X64.S
@@ -18,11 +18,11 @@ spill_registers:
     push %r13
     push %r14
     push %r15
-    // %rdi already contains the 'self' reference to the collector. We move the
-    // function pointer in %rsi to %r8 (a scratch register) so that the stack
-    // pointer can be placed in %rdi as the first arg slot.
-    mov %rsi, %r8
-    mov %rsp, %rsi
+    // %rdi contains the function pointer to our stack scanning code. We move it
+    // to %r8 (a scratch register) so that the stack pointer can be placed in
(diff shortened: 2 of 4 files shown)


ltratt:
Overall I like this PR, except possibly the title: I think it actually slightly increases the number of `unsafe` statements. I think the point here is that we stop seemingly safe code being able to undermine the guarantees that `unsafe` code needs to rely on?

jacob-hughes:
That's a good point. What we're really trying to do is wrap some of the unsafe pointer manipulation around a safe(er) API which has explicit lifetime requirements.

ltratt:
OK, when @vext01 is OK with this, maybe tweak the PR message and/or commit messages to bring this out a bit more?

vext01:
If I understand correctly, you use a reference to a `PhantomData` to add lifetimes to raw pointers. This seems sensible, but I'm unable to do a deep review of most of this code.

jacob-hughes:
> If I understand correctly, you use a reference to a `PhantomData` to add lifetimes to raw pointers. This seems sensible, but I'm unable to do a deep review of most of this code.

That's about the gist of it yeah.

ltratt:
@vext01 You're OK with the PR then?

@jacob-hughes If Edd is OK with it, then please force push any updates to commit messages / PR message and then shout. Thanks!

vext01:
Sorry, by my limited understanding, yes I'm happy.

jacob-hughes:
How about this? I've only changed the PR title and description, as I don't think the commits themselves were necessarily misleading.

ltratt:
Works for me!

bors r+

PR #33, opened 2020-01-21
Cache heap block pointers across a collection

jacob-hughes:
During marking, pointer identification is slow because it needs to
traverse a linked-list spanning the entire heap while looking for a
pointer. We now sort and cache block pointers in-use by the mutator when
a collection is triggered so that they can be looked up in O(logn) using
binary search.

Unfortunately this required quite a bit of code shuffling to keep the
borrow checker happy from the previous couple of PRs.

### DIFF
--- a/src/collector.rs
+++ b/src/collector.rs
@@ -88,9 +88,17 @@ impl DebugFlags {
     }
 }
 
+/// Holds information needed for the mark phase of GC
 pub(crate) struct MarkingCtxt<'mrk> {
+    /// The worklist holds references to heap blocks which were identified by
+    /// potential pointers somewhere in the object graph. During marking, the
+    /// entire worklist is traversed in order for a full transitive closure over
+    /// the object graph to be performed.
     pub(crate) worklist: GcVec<Block<'mrk>>,
+    /// The top (or start) of the current thread's stack.
     pub(crate) stack_start: usize,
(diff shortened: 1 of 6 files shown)


src/collector.rs
```diff
@@ -88,13 +73,115 @@ impl DebugFlags {
...
+    allocation_cache: AllocCache<'a>,
+    /// The data segment of an ELF binary contains statics. This section needs
+    /// scanning during marking as it could contain GC roots.
     pub(crate) data_segment_start: usize,
     pub(crate) data_segment_end: usize,
 }
 
+impl<'a> CollectionCtxt<'a> {
+    pub(crate) fn find(&mut self, word: usize) -> Option<Block<'a>> {
+        if !(word >= self.data_segment_end && word <= unsafe { HEAP_TOP }) {
```
ltratt: Why is `HEAP_TOP` surrounded by `unsafe`?

ltratt:
Let's see if you formatted things :)

bors try

ltratt:
> Let's see if you formatted things :)

You did -- excellent :)

jacob-hughes:
> You did -- excellent :)

Hooray for git hooks

src/collector.rs
jacob-hughes: It's a mutable static

ltratt: OK.

PR #34, opened 2020-01-23
Allow construction of uninitialized Gc values

jacob-hughes:
This lets us create a Gc value to a `Gc<MaybeUninit<T>>` with a new constructor:

```rust
pub fn new_from_layout<T>(layout: Layout) -> Option<Gc<T>> 
```

The constructor returns `None` if `layout` is less than `size_of::<T>()` because
this could accidentally trigger out-of-bounds accesses.

Since a `Gc<MaybeUninit<T>>` will not be dropped until it is initialized, we
also need to modify the sweep and finalization phases to make sure of this.

### DIFF
--- a/src/gc.rs
+++ b/src/gc.rs
@@ -2,8 +2,8 @@ use std::{
     alloc::{Alloc, Layout},
     any::Any,
     fmt,
-    marker::Unsize,
-    mem::{forget, transmute, ManuallyDrop},
+    marker::{PhantomData, Unsize},
+    mem::{forget, transmute, ManuallyDrop, MaybeUninit},
     ops::{CoerceUnsized, Deref, DerefMut},
     ptr::NonNull,
 };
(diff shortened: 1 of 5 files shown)


src/gc.rs
```diff
@@ -39,41 +39,99 @@ use crate::GC_ALLOCATOR;
...
         }
     }
+
+    /// Constructs a new `Gc<MaybeUninit<T>>` which is capable of storing data
+    /// up-to the size permissible by `layout`.
+    ///
+    /// This can be useful if you want to store a value with a custom layout,
+    /// but have the collector treat the value as if it were T.
+    ///
+    /// If `layout` is insufficiently sized to store `T`, a Gc value will not be
```
ltratt: The first sentence here is a bit too formal and hard to parse. Can we just say "`layout` must be at least as large as, and of the same alignment or bigger, as `T`." or similar? I then think we can remove the "this is because" sentence (in a sense it doesn't matter: the user *must* satisfy the contract, and "why" is often more detail than they need).

src/gc.rs
```diff
@@ -39,41 +39,99 @@ use crate::GC_ALLOCATOR;
...
+    /// allocated. This is because it's possible to accidentally access out-of-
+    /// bounds memory when treating the Gc value as if it contains a type larger
+    /// than its actual size.
+    pub fn new_from_layout(layout: Layout) -> Option<Gc<MaybeUninit<T>>> {
+        if layout.size() < Layout::new::<T>().size() {
```
ltratt: Need to check for alignment too.

src/gc.rs
```diff
@@ -39,41 +39,99 @@ use crate::GC_ALLOCATOR;
...
+    /// `T` available. This means that under-the-hood it must store a pointer to
+    /// `T`s drop method alongside its value.
+    ///
+    /// A `Gc<MaybeUninit<T>>` value will never call `T`'s `drop` method when it
+    /// is freed because the value may never have been initialized. However,
+    /// since `assume_init` promotes it to a `Gc<T>`, `T`'s `drop` method *must*
+    /// now be called during garbage-collection. For this to be possible, the
+    /// `drop` fn pointer must be extracted from T and stored alongside the
+    /// value, hence the run-time cost.
+    ///
```
ltratt: Random blank comment line.

src/gc.rs
```diff
@@ -39,41 +39,99 @@ use crate::GC_ALLOCATOR;
...
+    fn from_inner(ptr: NonNull<GcBox<T>>) -> Self {
+        Self {
+            ptr,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T> Gc<MaybeUninit<T>> {
+    /// Be aware that unlike with other smart-pointers such as
```
ltratt: Result of in person discussion: we'll probably remove this doc comment and replace it with something like "see [maybeuninit::assume_init]".

src/gc.rs
jacob-hughes: Fixed: 55aee509d1af3eb40a10836370a0123aefc8b1dc

src/gc.rs
jacob-hughes: Fixed: 93953995086dc4076967b69f5726ba4d29b14ffd

src/gc.rs
jacob-hughes: Fixed: 0024ec1b173a265f1a780822563fd2f823c91ab3

ltratt:
Please squash.

jacob-hughes:
Squashed.

PR #36, opened 2020-01-29
Fix use-after-free in preparation phase of collection

jacob-hughes:
After switching the collector to colour both GC and non-GC values during
marking, I forgot to update the preparation phase. This led to the
collector assuming some values' fields had already been traced. If those
fields were the only thing pointing to live values in the graph, then
those live values were erroneously freed.

### DIFF
--- a/gc_tests/tests/prep_phase_resets_non_gc_objects.rs
+++ b/gc_tests/tests/prep_phase_resets_non_gc_objects.rs
@@ -1,0 +1,24 @@
+// Run-time:
+//  status: success
+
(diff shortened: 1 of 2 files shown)


ltratt:
Can we add a test to stop this reoccuring?

jacob-hughes:
It's quite hard to test for this specifically without other co-founders perhaps being at play. I've added a test which can catch the most obvious cases.

ltratt:
Please squash.

jacob-hughes:
Squashed

PR #37, opened 2020-02-01
Add mechanism for specifying how the GC should trace values

jacob-hughes:
The collector currently treats all objects it encounters as opaque
blocks of possible pointer values. It then scans through these,
word-by-word, checking whether these possible pointers extend the object
graph. This approach is very simple, but it can miss pointers if they
are stored 'opaquely'. In other words, pointers which are stored in a
way which require some form of de-obfuscation before they can be
dereferenced will be missed by the collector.

To mitigate this, we need a way for the user to be able to tell the gc
that certain fields require de-obfuscating before they can be treated as
pointers. In this commit, we flesh out such an API.

The basic idea is to introduce a `Trace` trait for types to implement if they
contain opaque pointers. It defines as single `trace` method which
specifies how pointer fields can be de-obfuscated by the collector.

The `Trace` trait should not be implemented directly by the user,
however. Instead, we provide a custom derive attribute which we expect
users to use. Consider the following struct:
```rust
    struct S {
        opaque_ptr: usize,
        no_ptr: usize,
        normal_ptr: *mut Vec<...>
    }
```
If the collector traced `S` conservatively, it would miss this pointer
stored in the `opaque_ptr` field. To tell the collector how to
de-obfuscate it, we derive `Trace` and use a field level attribute as
shown below:
```rust
    #[derive(Trace)]
    struct S {
        #[trace = "path::to::deobfuscation::fn"]
        opaque_ptr: usize,
        no_ptr: usize,
        #[trace]
        normal_ptr: *mut Vec<...>
    }
```
The user needs to write a function to de-obfuscate each opaque field,
and then link it to the required field in the with the attribute:
`#[trace = "..."]`. The de-obfuscation function should take a single
argument, whose type is that of the field and return a pointer with the
type `*const u8`. For example, a de-obfuscation function for
`opaque_field` above might look like this:
```rust
    fn deobf_opaque_ptr(field: usize) -> *const u8 {
        (field | SOME_MASK) as *const u8
    }
```
Since a type which implements `Trace` will no longer be conservatively
traced, the user must also annotate other pointer-like fields which lead
to Gc values. This is done as above with the `#[trace]` attribute. This
is clearly error-prone and undesirable, and in the future this will not
be necessary.

This is insufficient to describe all types which may need a custom
tracing mechanism. However, I think it is quite ergonomic for the
common-case, and it can be built upon for more complex trace semantics.

The GC code required to respect this `Trace` API has not yet been
implemented.

### DIFF
--- a/src/trace.rs
+++ b/src/trace.rs
@@ -1,0 +1,67 @@
+#[derive(Debug, PartialEq)]
+pub enum TraceType {
+    Conservative,
+    Custom,
+}
+
+/// The trait used by the collector to determine how an object's fields should
+/// be traced.
+pub trait Trace {
+    fn trace(&self, _: &mut Vec<usize>) -> TraceType {
+        TraceType::Conservative
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use trace_derive::Trace;
+
+    #[test]
+    fn custom_trace() {
+        fn unmask_fn(x: usize) -> *const u8 {
+            (x + 1) as *const u8
+        }
+
+        #[derive(Trace)]
+        struct S {
+            #[trace = "unmask_fn"]
+            x: usize,
+        }
+
+        let s = S { x: 0 };
+        let mut trace_stack = Vec::<usize>::new();
+        let tt = s.trace(&mut trace_stack);
+        assert_eq!(tt, TraceType::Custom);
+        assert_eq!(trace_stack[0], 1);
+    }
+
+    #[test]
+    fn custom_trace_records_all_fields() {
+        fn unmask_fn(x: usize) -> *const u8 {
+            (x + 1) as *const u8
+        }
+
+        struct T;
+        #[derive(Trace)]
+        struct S<'a> {
+            #[trace = "unmask_fn"]
+            x: usize,
+            _y: *mut u8,
+            #[trace]
+            z: &'a T,
+        }
+
+        let t = T {};
+        let s = S {
+            x: 0,
+            _y: 123 as *mut u8,
+            z: &t,
+        };
+        let mut trace_stack = Vec::<usize>::new();
+        let tt = s.trace(&mut trace_stack);
+
+        assert_eq!(tt, TraceType::Custom);
+        assert_eq!(trace_stack, &[s.z as *const _ as usize, 1]);
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ packed_struct = "0.3"
 packed_struct_codegen = "0.3"
 parking_lot = { version = "0.10", features = ["nightly"] }
 static_assertions = "1.1"
+trace_derive = { path = "trace_derive" }
 
 [build-dependencies]
 rerun_except = "0.1"
@@ -23,3 +24,5 @@ tempdir = "0.3"
 name = "gc_tests"
 path = "gc_tests/run_tests.rs"
 harness = false
+
+[workspace]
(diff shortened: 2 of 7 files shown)


ltratt:
A possibly dumb question: what are the pros and cons of allowing users to specify an entire `trace` method vs. annotating individual members of a struct? I think it might be worth enumerating the major pros/cons before we move forwards.

jacob-hughes:
**Pros for annotating individual members of a struct:**
- Independent to the `trace` implementation. This is especially useful at these early stages where the precise rules and ergonomics of tracing are somewhat in flux, while I work out how it will interact with the collector.
- In a similar vein to above, I think that this might be easier to understand for non-gc people. All we are asking now is "provide me a function which turns this field into a legal pointer". It's familiar as it's a common pattern in serialization libraries.

**Cons of annotating individual members of a struct:**
- It doesn't cover all cases. A contrived example being that we could store an entire struct XORed with some key and it doesn't make sense for de-obfuscation functions to be on a field level.
- Potential of requiring user to define many helper functions if lots of fields need unmasking.

Realistically though, I don't see why we can't have our cake and eat it. I can see a mix of both being useful: annotated fields for the common-case but with custom trace implementations if you're doing more complex stuff like treating an entire block like an object of a different type (i.e. natrob). 

My thoughts are that we should first explore the field annotation approach as no doubt our `trace` implementation will require some tweaking down the road. This approach lets us better isolate those changes from applications using gcmalloc as a library. That's not to say I think that gcmalloc is ready to go on crates.io and be used by the wider-world, but it will certainly make development and testing a little less painful.

src/trace.rs
```diff
@@ -0,0 +1,67 @@
+#[derive(Debug, PartialEq)]
+pub enum TraceType {
+    Conservative,
+    Custom,
+}
+
+/// The trait used by the collector to determine how an object's fields should
+/// be traced.
+pub trait Trace {
+    fn trace(&self, _: &mut Vec<usize>) -> TraceType {
```
ltratt: Why is the argument `&mut Vec<usize>`? Should it be `&[u8]`?

jacob-hughes: Because we push pointers into it inside the `trace` method.

ltratt: Ah!

ltratt:
OK, I think you've convinced me that, at the very least, this is a reasonable route to take for the immediate future. OK to merge?

jacob-hughes:
Cool! Yes please.

ltratt:
That looks like a genuine test failure (a reordered list)?

jacob-hughes:
Interesting. I'll have a look.

jacob-hughes:
This is a strange case of "works on my machine". There seems to be some non-determinism in the order in which the CTMP macro traverses fields. We don't particularly care about the order, so we could just sort the lists before we compare them. But it might be worth me investigating why, as it's a little odd.

ltratt:
Yes, I think we should nail this down.

jacob-hughes:
Test failure should be fixed with: 99f96d321f3b83c37f07fbf3d7d55743ab04282b

There is no non-determinism with `trace` field traversal as I had first suspected. Instead, on old versions of nightly, assignments such as `let t = &T{}` would lead to `t`'s reference equaling the constant `1`. Since, by coincidence, the other field I was comparing also happened to be `1`, I missed an out-of-order bug in my assertion locally!

ltratt:
I would not have guessed that! Please squash.

jacob-hughes:
Squashed

jacob-hughes:
Erm.. That's an interesting one.

jacob-hughes:
I'm guessing some APIs have moved around in nightly 02-02-2020. I'll update my imports.

jacob-hughes:
Fixed usages of `Alloc` to use the new `AllocRef` API in rustc. No need to squash.

PR #38, opened 2020-02-17
Add lang_tester framework for Miri tests

jacob-hughes:
Unfortunately this is a bit brittle because of limitations of the
cargo-miri binary [1]. To mitigate this, we create a temp directory and
rsync over the src files to compile a separate gcmalloc target with miri
flags. We then copy over the tests one at a time in order to run them
under Miri.

[1]: https://github.com/rust-lang/miri/issues/1173

### DIFF
--- a/gc_tests/miri_test_setup.sh
+++ b/gc_tests/miri_test_setup.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -e;
+
+PROJECTDIR=$1
+TMPDIR=$2
+
+# Miri deps cause breakages when used with non-miri targets. So we copy gcmalloc
+# sources over to a tempdir to compile with miri flags so we can nuke it
+# afterwards.
+rsync -a --partial --info=progress2 --exclude .git --exclude target --exclude gc_tests $PROJECTDIR/ $TMPDIR
+
+mkdir -p "$TMPDIR/src/bin/"
+
+# Running this dummy binary forces cargo-miri to compile gcmalloc with miri
+# flags as a dependency.
+echo "extern crate gcmalloc; fn main() {}" > $TMPDIR/src/bin/miri.rs
+(cd $TMPDIR && cargo-miri miri setup)
+(cd $TMPDIR && cargo-miri miri run -- -Zmiri-ignore-leaks)
+
+rm $TMPDIR/src/bin/miri.rs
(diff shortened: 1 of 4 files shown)


gc_tests/miri_test_setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -e;
+
+PROJECTDIR=$1
+TMPDIR=$2
+
+# Miri deps cause breakages when used with non-miri targets. So we copy gcmalloc
+# sources over to a tempdir to compile with miri flags so we can nuke it
+# afterwards.
+rsync -a --partial --info=progress2 --exclude .git --exclude target --exclude gc_tests $PROJECTDIR/ $TMPDIR
```
ltratt: `rsync`?!

gc_tests/run_miri_tests.rs
```diff
@@ -0,0 +1,78 @@
...
+    let projd = proj_dir();
+    let testsd = test_dir();
+
+    let tmpd = TempDir::new("miri_tests").unwrap();
+    let mut binsd = PathBuf::new();
+    binsd.push(tmpd.path());
+    binsd.push("src");
+    binsd.push("bin");
+
+    let mut setup = Command::new("./gc_tests/miri_test_setup.sh")
```
ltratt: I wondered if it would be better to inline what `miri_test_setup.sh` does into Rust? But perhaps that's overkill.

gc_tests/miri_test_setup.sh
jacob-hughes: This could probably be `cp` now. I used `rsync` while developing because there was a lot of trial-and-error testing and I wanted to speed up transfer of targets. I'm curious, is the "?!" a comment on the overkill or poor platform compatibility. If the latter, I'm afraid there's plenty more where that came from with this approach to getting Miri running.

ltratt: Yes, `cp` is more portable than `rsync`, and given that we copy to a fresh directory each time, I think `cp` is sufficient.

jacob-hughes:
There is a subtle bug with this where things break occasionally when compiled alongside the other `gc_tests`. It will take some time to investigate but since this isn't going to be used just yet I am going to put it on hold.

PR #39, opened 2020-05-26
Replace tempdir with tempfile

jacob-hughes:
tempdir is deprecated and presents as a security vulnerability in cargo
audit.


jacob-hughes:
I think this is failing because rustfmt doesn't work on the current version of nightly. I'll tweak the build script to `cargo fmt` conditionally.

ltratt:
OK.

jacob-hughes:
bors try

jacob-hughes:
That seems to have done the trick

